PR #3, opened 2019-06-07
Add a bibliography link to the home page

daly:
Create a common source of bibliographic references for community use.


daly:
This creates a new file "bibliography.tex" and provides a link on the _includes/header.html home page

robertylewis:
Tim, thanks for the PR. I would rather see this as a formatted HTML file, instead of (or in addition to) a raw .bib file. I also think it makes sense to focus on community papers in this list, since the official Lean publications are on the main leanprover site, although we could include both in the .bib file.

robertylewis:
Merged.

PR #9, opened 2020-05-11
Changed dropdown menus to activate on hover to get rid of popper.js

Steffan153:
I changed the dropdown menus so that they'll appear when you hover instead of clicking, and therefore removed popper.js.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "editor.formatOnSave": false
-}
--- a/css/lean.css
+++ b/css/lean.css
@@ -1,4 +1,13 @@
 @charset "UTF-8";
+
+li.nav-item.dropdown:hover .dropdown-menu {
+  display: block;
+}
+
+.dropdown-menu {
+  margin-top: -1px !important; /* When the dropdown comes down on hover, this will make it easier to get to the dropdown contents without losing it. */
+}
+
 /*!
  * Bootstrap v4.4.1 (https://getbootstrap.com/)
  * Copyright 2011-2019 The Bootstrap Authors
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -61,7 +61,6 @@
 	  {% endblock %}
   </div>
     <script src="https://code.jquery.com/jquery-3.4.1.slim.min.js" integrity="sha384-J6qa4849blE2+poT4WnyKhv5vZF5SrPo0iEjwBvKU7imGFAV0wwj1yYfoRSJoZ+n" crossorigin="anonymous"></script>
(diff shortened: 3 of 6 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
bryangingechen, line 1: Remove this and add `.vscode/*` to the `.gitignore`.

css/lean.css
```diff
@@ -1,4 +1,13 @@
 @charset "UTF-8";
+
+li.nav-item.dropdown:hover .dropdown-menu {
+  display: block;
+}
+
+.dropdown-menu {
+  margin-top: -1px !important; /* When the dropdown comes down on hover, this will make it easier to get to the dropdown contents without losing it. */
```
bryangingechen, line 8: Is the `!important` necessary?

templates/_base.html
```diff
@@ -61,8 +61,8 @@
 	  {% endblock %}
   </div>
     <script src="https://code.jquery.com/jquery-3.4.1.slim.min.js" integrity="sha384-J6qa4849blE2+poT4WnyKhv5vZF5SrPo0iEjwBvKU7imGFAV0wwj1yYfoRSJoZ+n" crossorigin="anonymous"></script>
-    <script src="https://cdn.jsdelivr.net/npm/popper.js@1.16.0/dist/umd/popper.min.js" integrity="sha384-Q6E9RHvbIyZFJoft+2mJbHaEWldlvI9IOYy5n3zV9zzTtmI3UksdQRVvoxMfooAo" crossorigin="anonymous"></script>
-	<script src="{{ base_url }}/js/bootstrap.min.js"></script>
+    <!-- <script src="https://cdn.jsdelivr.net/npm/popper.js@1.16.0/dist/umd/popper.min.js" integrity="sha384-Q6E9RHvbIyZFJoft+2mJbHaEWldlvI9IOYy5n3zV9zzTtmI3UksdQRVvoxMfooAo" crossorigin="anonymous"></script> -->
```
bryangingechen, line 64:
```suggestion
```

css/lean.css
```diff
@@ -1,4 +1,13 @@
 @charset "UTF-8";
+
```
bryangingechen, line 2: These changes should be made to `scss/lean.scss` file, since `css/lean.css` should be generated by `sass`: see https://github.com/leanprover-community/leanprover-community.github.io#building

css/lean.css
Steffan153: yep

Steffan153:
@bryangingechen All done, please check again.

css/lean.css
bryangingechen: Sorry, by "necessary", I don't mean "does it fail to work if you remove `!important` here?". I mean: are you sure there is no way to get the desired effect without using it?

(I'm far from a CSS expert, so I'm just recalling recommendations like [this one](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) to avoid `!important` in most cases if possible.)

Steffan153: Bootstrap sets a margin-top by default, so I have to override it.

There might be a way to avoid it, but CSS order of rules is a big mess and I don't like to mess with it.

bryangingechen: Yes, I just had a look and you're right. Thanks.

bryangingechen:
Thank you! Note that I'll make one more tweak after I merge this (mainly changing `lean.css` so that it is precisely what is generated by `sass` from `lean.scss`).

PR #10, opened 2020-05-13
Some pointers to docs

jcommelin:
They are not all completely uncontroversial. So I made a PR.


data/overview.yaml
```diff
@@ -13,9 +13,9 @@ group:
     orbit: 'group_theory/group_action.html#mul_action.orbit'
     quotient space: 'group_theory/group_action.html#mul_action.orbit_equiv_quotient_stabilizer'
     class formula: ''
-    conjugacy classes: ''
+    conjugacy classes: 'algebra/group/conj.html'
```
jcommelin, line 16: I'm not sure if this is really conjugacy *classes*. But we have all the ingredients.

data/overview.yaml
```diff
@@ -49,12 +49,12 @@ linear algebra:
   kernel of a linear map: 'linear_algebra/basic.html#linear_map.ker'
   quotient space: 'linear_algebra/basic.html#submodule.quotient'
   sum of subspaces: 'linear_algebra/basic.html#submodule.complete_lattice'
-  direct sum: ''
+  direct sum: 'data/finsupp.html#finsupp.vector_space'
```
jcommelin, line 52: This is direct sums... but not dependent. And `dfinsupp` seems to be gone. Alternatively, we could link to the vector space structure on `V × W`.

data/overview.yaml
PatrickMassot: I don't think we should cheat in this file. Leaving this line empty creates an incentive to actually fix this tiny gap. We can even hope that the person fixing this gap will use the momentum to add some more.

data/overview.yaml
PatrickMassot: What is intended here is clearly the bigop $\bigoplus V_i$.

data/overview.yaml
jcommelin, line 16:
```suggestion
    conjugacy classes: ''
```

data/overview.yaml
jcommelin, line 52:
```suggestion
  direct sum: ''
```

PR #11, opened 2020-05-14
make markdown headings linkable

bryangingechen:
This PR makes markdown headings linkable like the way they are on github. Currently it's set up to look like this (adding a `#` symbol to the left):

Screenshot 2020-05-13 22 32 32

If this is too ugly, it shouldn't be too hard to change the template.

### DIFF
--- a/css/lean.css
+++ b/css/lean.css
@@ -9997,7 +9997,3 @@ li.nav-item.dropdown:hover .dropdown-menu {
 .markdown-heading:hover > .hover-link {
   display: inline;
 }
-
-.hover-link:hover {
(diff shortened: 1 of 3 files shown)


PatrickMassot:
Would it be hard to have it GitHub style, with an invisible link symbol that shows up on hover?

bryangingechen:
I'll give it a shot.

bryangingechen:
Now the `#` only appears when you hover over the heading:

Screenshot 2020-05-14 22 14 04

PatrickMassot:
Ok, let's try

PR #13, opened 2020-05-15
add "Suggest edits to this page on GitHub" to bottom of markdown pages

bryangingechen:
This adds a navbar to the bottom of all pages generated from markdown with "Edit this page on GitHub" in the bottom right:

Screenshot 2020-05-14 23 25 57

cf. https://github.com/leanprover-community/mathlib/pull/2676#issuecomment-628487986 by @robertylewis

### DIFF
--- a/css/lean.css
+++ b/css/lean.css
@@ -9973,6 +9973,20 @@ h3 {
     border-color: #dee2e6;
   }
 }
+html {
+  height: 100%;
+}
+
+body {
+  min-height: 100%;
+  display: flex;
+  flex-direction: column;
+}
+
+body > #wrapper {
+  flex-grow: 1;
+}
+
 code {
   color: #931e53;
 }
--- a/make_site.py
+++ b/make_site.py
@@ -174,7 +174,7 @@ def render_site(target: Path, base_url: str, reloader=False):
 
     def get_contents(template):
         return { 'content': Path(template.filename).read_text(encoding='utf-8').replace('img/',
-            base_url+'/img/') }
+            base_url+'/img/'), 'name': template.name }
 
     def url(raw: str):
         return raw if raw.startswith('http') else base_url + raw
--- a/scss/lean.scss
+++ b/scss/lean.scss
@@ -2,6 +2,22 @@ $enable-gradients:                            true;
 $enable-shadows : true;
 @import "../bootstrap-4.4.1/scss/bootstrap";
 
+// push the footer to the bottom of the screen
+// https://stackoverflow.com/a/54100290
+html {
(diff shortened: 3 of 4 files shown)


robertylewis:
Would it be better to phrase it as "suggest edits to this page" or something like that? (To make it clear that for most people, the changes have to be approved?)

bryangingechen:
@robertylewis Done.

robertylewis:
I like it, but it doesn't seem quite right:
* It shows up on the home page, which isn't generated (directly) from markdown. The link goes to the templates directory.
* It doesn't show up on the mathlib-overview page, which is direct from markdown.
* On pages that don't take my full vertical screen, the bar comes immediately after the page body, with blank space until the bottom. Could we float it to the bottom of the screen?

bryangingechen:
> It shows up on the home page, which isn't generated (directly) from markdown. The link goes to the templates directory.

Good catch! I think I've fixed this.

> It doesn't show up on the mathlib-overview page, which is direct from markdown.

Strange, I can't reproduce this.

> On pages that don't take my full vertical screen, the bar comes immediately after the page body, with blank space until the bottom. Could we float it to the bottom of the screen?

I think I got this working.

robertylewis:
Oh, I see. The link in the upper left is hardcoded to the real site, I got off my local version without realizing it.

Home page looks fixed. Seeing the bar on all the markdown pages. But it's not floated. I don't think this is a big deal, if it's too much trouble we can ignore it.

bryangingechen:
> Seeing the bar on all the markdown pages. But it's not floated.

What do you mean by "floated"? Did you want it to be pinned to the bottom of the viewport as opposed to the bottom of the page content?

robertylewis:
Sorry. I know nothing about web design or the terminology behind it, if you can't tell! It just looks weird to have the footer bar in the middle of my screen. I think it should be at the bottom of the page content, if the page content fills more than the viewport, or otherwise at the bottom of the viewport. Does that make any sense?

bryangingechen:
Oh, I thought I fixed that. At least it works in all three of Firefox, Chrome and Safari on my mac...

robertylewis:
Okay, this is confusing. I'm on 15e9161 and it displays like your first picture, in both Chrome and Firefox. I've refreshed the cache (and never loaded this site in Firefox anyway).

robertylewis:
OH. Never mind, I'm dumb, did that build without --local. Looks great!

robertylewis:
I'll let @PatrickMassot approve before merging.

PatrickMassot:
Thanks!

PR #19, opened 2020-05-15
[WIP] Add a nice rendered page for 100 theorems

alexjbest:
I started trying to make a nice rendered page for the 100 theorems list as I didn't see one

current issues (please find more problems):

 - [x] List of links don't work well I think we should specify the format more, maybe use a list, even if there is just one

    ```
    links:
       - https://example.fds
    ```

- [ ] when decl is just a name without a full path it doesn't work (fixed data)

- [x] lists of decls don't work

- [x] an automatic total count would be nice

- [ ] I changed the make_site to explicitly say python3 (does this cause breakage for anyone) 

- [ ] the page should be linked to from the main site (no, it shouldn't)

- [x] clarify the authorship of the original list

### DIFF
--- a/templates/100.html
+++ b/templates/100.html
@@ -14,7 +14,7 @@
     <div class="list-group-item">
         <h5 class="card-title markdown-heading" id="{{ theorem.number }}">{{ theorem.number }}: {{ theorem.title }} <a class="hover-link" href="#{ theorem.number }">#</a></h5>
         <h6 class="card-title">Author: {{ theorem.author }}</h6>
-        {% if theorem.decl %}<p><a href="/docs/{{ theorem.decl}}">{{ theorem.decl }}</a></p>{% endif %}
+        {% if theorem.decl %}<p><a href="/mathlib_docs/{{ theorem.decl}}">{{ theorem.decl }}</a></p>{% endif %}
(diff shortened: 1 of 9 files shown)


templates/100.html
```diff
@@ -0,0 +1,24 @@
...
+        <h1>100 theorems</h1>
+        
+        <p><a href="https://www.cs.ru.nl/~freek/">Freek Wiedijk</a> has created <a href="https://www.cs.ru.nl/~freek/100/">a list</a> of 100 theorems as targets for formalization as a way of tracking and comparing prominent theorem provers.
+        The progress on problems on this list in Lean is as follows:</p>
+
+  <div class="list-group">
+    {% for theorem in hundred_theorems %}
+    {% if theorem.link or theorem.decl or theorem.author or theorem.note %}
+    <div class="list-group-item">
+        <h5 class="card-title">{{ theorem.number }}: {{ theorem.title }}</h5><h6 class="card-title">Author: {{ theorem.author }}</h5>
```
bryangingechen, line 15:
```suggestion
        <h5 class="card-title markdown-heading" id="{{ theorem.number }}">{{ theorem.number }}: {{ theorem.title }} <a class="hover-link" href="#{ theorem.number }">#</a></h5>
        <h6 class="card-title">Author: {{ theorem.author }}</h6>
```
This should make the theorems linkable (following #11). I also fixed what looks to be a typo: `</h5>` -> `</h6>`

PR #20, opened 2020-05-15
Specify some attributions more

alexjbest:


data/100.yaml
```diff
@@ -58,7 +58,7 @@
    17:
       - title  : De Moivre’s Theorem
       - decl   : data/complex/exponential.html#complex.cos_add_sin_mul_I_pow
-      - author : mathlib
+      - author : Chris Hughes and Abhimanyu Pallavi Sudhir
```
bryangingechen, line 61: I don't know how we are picking authors, but from what I can tell, the theorem was originally added as `pow_of_cos_add_I_mul_sin` by @abhimanyupallavisudhir in https://github.com/leanprover-community/mathlib/pull/483/commits/158d0c2d60d55b3700575618035fbd9ad40da1f1#diff-689fa226d40ff8b8b594616209af8bcaR564 and then renamed and golfed by @kckennylau in https://github.com/leanprover-community/mathlib/pull/483/commits/09c92764b80d8ee417d0a77bf756145b4abd4698#diff-689fa226d40ff8b8b594616209af8bcaR574. (@ChrisHughes24 reviewed the PR(s) #483 and #745.)

alexjbest: I figured @abhimanyupallavisudhir proved the theorem itself but @ChrisHughes24 did a lot of the results immediately above / theory that is used so I erred on the side of giving more credit out than less is all ;)

ChrisHughes24: I guess I did do a lot of the work so I suppose I deserve some credit.

PR #44, opened 2020-05-30
overview: Section 6

PatrickMassot:
I went through Section 6. It also displays what I think we should do. In particular:
* I moved stuff around a bit, to improve consistency
* I split a couple of items
* when an item has several targets but finding one tells you about the other, I didn't split, and indicated where to find one target. For instance, "Circular trigonometric function" links to the definition of sin. I did split circular and hyperbolic, especially since we don't have reciprocals hyperbolic trig functions
* When we can obviously express something but the definitions is not there, I didn't write anything. I think we should add those definitions to mathlib, in particular for the sake of discoverability. For instance we don't have convergent series of real numbers.

What I think I have a hard time to do is being consistent about the following question: what do we do when we have a very general form of something, but we would need an API for the elementary version? For instance I resisted the temptation to write that we have Taylor series. We have them in several variables. But since we have the `deriv` vs `fderiv` distinction, I say we don't have Taylor series in one variable.

### DIFF
--- a/data/overview.yaml
+++ b/data/overview.yaml
@@ -270,62 +270,66 @@ Affine spaces and associated vector spaces:
 # 6.
 Single Variable Real Analysis:
   Real Numbers:
-    $\R$ as a field: 'data/real/basic.html#real.division_ring'
-    Topology of R: 
-    Additive subgroups of $\R$:
-    Series of real numbers:
-      convergence:
-      limit point:
-      recurrent sequences:
-      Limit infimum and supremum:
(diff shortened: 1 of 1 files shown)


jcommelin:
In my PR I didn't split "Z/nZ and its units" but linked to a theorem about `a : zmod n` being a unit iff `a` is coprime to `n`. This way, people can still easily discover `zmod n`. But of course we could simply add an entry for `zmod`.

PR #45, opened 2020-05-30
feature (cleanups in overview.yaml)

paulvanwamelen:
This needs more work, but seems like lot's of people are working on this, so didn't want to wait for it to be "perfect".

### DIFF
--- a/data/overview.yaml
+++ b/data/overview.yaml
@@ -1,33 +1,33 @@
 # Data gathered from http://media.devenirenseignant.gouv.fr/file/agreg_externe/59/7/p2020_agreg_ext_maths_1107597.pdf
 # 1.
-Linear algebra:
-  vector space: 'algebra/module.html#vector_space'
-  linear map: 'algebra/module.html#linear_map'
-  product of vector spaces: 'algebra/pi_instances.html#prod.module'
-  vector subspace: 'algebra/module.html#subspace'
-  range of a linear map: 'linear_algebra/basic.html#linear_map.range'
-  kernel of a linear map: 'linear_algebra/basic.html#linear_map.ker'
-  quotient space: 'linear_algebra/basic.html#submodule.quotient'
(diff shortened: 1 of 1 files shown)


data/overview.yaml
```diff
@@ -115,30 +115,30 @@ Groups:
...
+  Divisibility in integral domains: ''
   Irreducible elements: 'algebra/associated.html#irreducible' 
   Invertible elements: 'algebra/invertible.html#invertible'
   Coprime elements: 'ring_theory/ideals.html#ideal.is_coprime'
   Unique factorisation domain (UFD): 'ring_theory/unique_factorization_domain.html'
   Greatest common divisor: 'algebra/gcd_domain.html#gcd_domain.gcd'
   Least common multiple: 'algebra/gcd_domain.html#gcd_domain.lcm'
   $A[X]$ is a UFD when $A$ is a UFD: ''
-  Principal ring: 'ring_theory/principal_ideal_domain.html#submodule.is_principal'
+  Principal ideal domain: 'ring_theory/principal_ideal_domain.html#submodule.is_principal'
```
bryangingechen, line 141: Note that this was previously discussed [here](https://github.com/leanprover-community/leanprover-community.github.io/pull/27#discussion_r429595741).

bryangingechen:
Can you explain why you capitalized some items and uncapitalized others? If there's a rule here, we should give it explicitly somewhere, so maybe we need to start compiling a `overview_style.md` file.

paulvanwamelen:
I though the top level things should be "all capitals". The next level should start with a capital. And the third level should be all lower case.
At least that looked like what the intent was.
But then it got messy because in places there is a fourth level and some top level things don''t have a middle level.
Also the original document has this inconsistency.

PatrickMassot:
I'm working on fixing levels. But I agree it's nicer to have capitals for headers and lower cases for leaves. We could capitalize in the building script, but having a nice YaML source is also a worthy goal.

PatrickMassot:
Thanks!

PR #46, opened 2020-05-31
clarify scope of `leanproject get`

jalex-stark:


PatrickMassot:
I don't think you need a `src` folder. Also, I'm not sure why you explain half of the magic. There are two levels of abbreviation. You can use `GitHub_user_name/repo_name` or only `repo_name` if the user is `leanprover-community`, or a full url if you are not using GitHub at all.

templates/install/project.md
```diff
@@ -59,7 +59,7 @@ Open a terminal.
 
 * Go the the directory where you would like this package to live.
 
-* Run `leanproject get tutorials`.
+* Run `leanproject get tutorials`. (This is just a shortcut for `leanproject get https://github.com/leanprover-community/tutorials.git`. You can use `leanproject get <git-url>` on any project which has a top-level file `leanpkg.toml` and top-level directory `src/`.)
```
jalex-stark, line 63:
```suggestion
(This is just.a shortcut for `leanproject get leanprover-community/tutorials`, which is again a shortcut for `leanproject get https://github.com/leanprover-community/tutorials.git`. You can use `leanproject get <git-url>` on any git repo and `leanproject get <github-username>/< project-name>` for any GitHub repo. This will succeed as long as the project has a top-level file `leanpkg.toml`.)
```

templates/install/project.md
```diff
@@ -59,7 +59,7 @@ Open a terminal.
...
```
jalex-stark, line 62:
```suggestion
* Run `leanproject get tutorials`. (This is just.a shortcut for `leanproject get leanprover-community/tutorials`, which is again a shortcut for `leanproject get https://github.com/leanprover-community/tutorials.git`. You can use `leanproject get <git-url>` on any git repo and `leanproject get <github-username>/< project-name>` for any GitHub repo. This will succeed as long as the project has a top-level file `leanpkg.toml`.)
```

jalex-stark:
> I don't think you need a `src` folder. Also, I'm not sure why you explain half of the magic. There are two levels of abbreviation. You can use `GitHub_user_name/repo_name` or only `repo_name` if the user is `leanprover-community`, or a full url if you are not using GitHub at all.

Thanks, I think I implemented the suggestion that follows from your comment.

I suppose I should have said this at the beginning: the purpose here is to dissolve the misconception that `leanproject get` works only on GitHub projects owned by leanprover-community. (In fact I had for a long time the worse misconception that it worked only for some list of projects hardcoded in the source code of `leanproject`.)

PatrickMassot:
I think the comment is now accurate. But I fear it clutters this page. Maybe we could simply write "(see more information [about `leanproject`](https://github.com/leanprover-community/mathlib-tools/blob/master/README.md))". I'd like to have more opinion here.

Also, this web page is among the very few that are duplicated on GitHub and on the community website, so any update should be propagated there.

PR #49, opened 2020-06-01
Update formalizations.yaml

cipher1024:


PatrickMassot:
Nice, it's a very good idea to include also something on the programming language side. You need a url though. And indeed the build failed. Once this is fixed, we should also check how the layout behaves. Do we get four columns or are there two rows of two columns? I don't remember how I set that up.

cipher1024:
I'm not clear on how this works either. I'm going to follow your lead here.

bryangingechen:
I just did a local build of this branch and it looks like we get 4 columns, until the width of the page is less than or equal to 575 px. The 4 columns do become a bit too narrow when the width of the page is less than 800 px or so, but I guess the same thing is true (to a lesser extent) with the current 3 column setup.

PatrickMassot:
Thanks Bryan! Simon I'm really sorry I forgot this PR. Actually the current layout is pretty bad on intermediate sized screens, I clearly didn't check this carefully enough.

Bryan, could you have a look at the version I just pushed in this branch? I switched to two columns and a setup that could easily be changed further if we want to switch to three columns when we reach six cards. I guess that people who will want to be the fifth card will need to find friends to make it six, for layout purposes.

bryangingechen:
@PatrickMassot Looks great now! Feel free to merge if you're also happy.

PR #50, opened 2020-06-02
bilinear and quadratic forms

kbuzzard:
I don't know how to check to see if I made any typos.

I think this might be all we have.


PatrickMassot:
Thanks!

PR #54, opened 2020-06-06
print types in 100 theorems list

robertylewis:
image

This is WIP, but posting to get comments. The doc page now generates a map from declaration name to location including type information. So we can use this map to display more info on the 100 theorems page.

Some of the links in the YAML file are broken already. I've commented these out, didn't investigate where they moved to.

TODO:
* It doesn't display when there are multiple declarations right now. This is easy enough, just takes a little restructuring of the data structure.
* Plenty of CSS/design work is needed.
* YAML entries no longer need the full path. I'm stripping everything before `#`. We can remove the unneeded info.
* Entries that are not in mathlib should have the type written explicitly to match.

If anyone would like to adopt this I'm happy to hand it over.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
       - 'master'
       - 'oldsite'
   schedule:
-    cron: '0 4 * * *'
+    - cron: '0 4 * * *'
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
     branches-ignore:
       - 'master'
       - 'oldsite'
+  schedule:
+    cron: '0 4 * * *'
 
 jobs:
   build:
--- a/css/lean.css
+++ b/css/lean.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /*!
  * Bootstrap v4.4.1 (https://getbootstrap.com/)
  * Copyright 2011-2019 The Bootstrap Authors
@@ -32,21 +33,25 @@
   --breakpoint-lg: 992px;
   --breakpoint-xl: 1200px;
   --font-family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
-  --font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace; }
+  --font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
+}
 
 *,
 *::before,
 *::after {
-  box-sizing: border-box; }
+  box-sizing: border-box;
+}
 
 html {
   font-family: sans-serif;
   line-height: 1.15;
   -webkit-text-size-adjust: 100%;
-  -webkit-tap-highlight-color: rgba(0, 0, 0, 0); }
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+}
 
 article, aside, figcaption, figure, footer, header, hgroup, main, nav, section {
-  display: block; }
+  display: block;
+}
 
 body {
   margin: 0;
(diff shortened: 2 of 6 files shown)


robertylewis:
Also note that the same map can be used for the mathlib overview. Even if you don't want to display types, you can find up-to-date links.

robertylewis:
image

The styling is now slightly better, and it handles entries with multiple declarations.

robertylewis:
Note, I didn't touch `lean.css` by hand, just the .scss file. I guess the large diff is because of a different version of scss or something?

jcommelin:
Thanks for doing this. @bryangingechen shall we merge this?

bryangingechen:
There are a few points that we'll have to address sooner or later. I think at least the last one should be figured out before we merge:
- where did the commented-out links go?
- what version of SCSS should we be using?
- this PR makes `100.html` dependent on `mathlib_docs`. Should we be triggering a website rebuild every time we rebuild the docs?

robertylewis:
> * this PR makes `100.html` dependent on `mathlib_docs`. Should we be triggering a website rebuild every time we rebuild the docs?

No more so than before. The links to source will be stable, since they go to specific versions of Lean/mathlib on GitHub. Links to the docs will be at least as stable as the current setup -- they'll work unless the declarations move to a new file, just like the current ones.

I think rebuilding the website once daily might be good, but triggering it from mathlib_docs seems excessive.

robertylewis:
Before merging this, we should remove the redundant paths from the yaml and update the corresponding code.

bryangingechen:
> > * this PR makes `100.html` dependent on `mathlib_docs`. Should we be triggering a website rebuild every time we rebuild the docs?
> 
> No more so than before. The links to source will be stable, since they go to specific versions of Lean/mathlib on GitHub. Links to the docs will be at least as stable as the current setup -- they'll work unless the declarations move to a new file, just like the current ones.
> 
> I think rebuilding the website once daily might be good, but triggering it from mathlib_docs seems excessive.

Makes sense. Do you mind adding a daily cron job to the actions workflow?

bryangingechen:
Also, do we want to switch to force-pushing to `master`? We can do that in a later PR too.

robertylewis:
> * where did the commented-out links go?

Based on [this](https://github.com/leanprover-community/mathlib/blob/100-thms/docs/100-theorems.md#72) the Sylow ones were never in mathlib to begin with. Don't ask [me](https://github.com/leanprover-community/leanprover-community.github.io/blame/61234e652862543bb08eea646f3f07031e1e34c4/data/100.yaml#L217) why they got added to the list as mathlib links...

robertylewis:
@PatrickMassot what do you think about this? If you approve, I'll change the yaml file to get rid of the redundant parts of the URLs. I can also remove my css changes and let someone else regenerate from the scss; does the CI do that when deploying the site?

robertylewis:
I changed the error handling to print a warning for broken links. But note that this can be used as CI for the 100 theorems list if it throws an exception. #65 fixes two problems that showed up in the previous commit.

PatrickMassot:
I haven't followed the discussion here since I have very little interest for this list. And I don't know anything about the CI aspects of all this. So I'm not sure why I'm pinged here. I certainly don't mind merging this.

bryangingechen:
Maybe this is for a separate PR, but it would be good to add CI warnings / errors for `overview.yaml` as well. (Are there other YAML files that reference mathlib?)

Currently the CI isn't doing anything with SCSS or CSS, but it shouldn't be hard to add a step for that. In that case I guess we'd want to delete the generated CSS from the repo, but that would mean that people would have to run scss to build the site locally.

robertylewis:
Yes, `overview.yaml` should use the same linkification map that I'm adding in this PR. 

@PatrickMassot I tagged you because I think of you as "in charge" of the website ;) Thought you might have an opinion about adding the dependency on the `mathlib_docs` map.

I'll finish cleaning this up and get it ready to merge. It would be great if someone else could run scss so the css diff isn't so unnecessarily large.

robertylewis:
I tried regenerating the css on my office computer and it's the same as at home.

bryangingechen:
I've pushed a commit with the new generated CSS but the diff is still huge. It looks like your lean.css file was generated with Ruby `sass` with the default "nested" output style: https://sass-lang.com/documentation/cli/ruby-sass#style and in fact looking at the history, it looks like the CSS has been that way since 8fba9640.

Before that commit and in the commit I just pushed, I generated the CSS file with dart `sass`, which has a different default "expanded" output style: https://sass-lang.com/documentation/cli/dart-sass#style

This doesn't explain the large diff between your CSS and the one from 8fba9640 though, maybe there are differences in the output of Ruby `sass`?

PatrickMassot:
The only correct solution is to generate CSS from SCSS as part of the site building, and remove the CSS file from the git repository. I put it originally because I feared some people might want to tweak the website without having sass. Now the website is more stable, and it turned out there aren't so many people working on it.

bryangingechen:
> The only correct solution is to generate CSS from SCSS as part of the site building, and remove the CSS file from the git repository.

Sounds good. I'll try to do this in another PR this weekend.

@robertylewis if you're happy with this, you can go ahead and merge.

PR #55, opened 2020-06-07
add Google authentication

robertylewis:
This makes Google happier about indexing and lets us see some basic stats.

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -289,6 +289,8 @@ def render_site(target: Path, base_url: str, reloader=False):
 
     for folder in ['css', 'js', 'img', 'papers']:
         subprocess.call(['rsync', '-a', folder, str(target).rstrip('/')])
+    subprocess.call(['rsync', '-a', 'googlef0c00cb4d31b246f.html', str(target).rstrip('/')])
+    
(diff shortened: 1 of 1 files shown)


robertylewis:
If we want more interesting stats, we could add analytics tracking to each page. Interesting or invasive?

jcommelin:
I would like to avoid google analytics. I'm hoping that once the student project on the search bar is done, we could get rid of all ties to google.

robertylewis:
That's fair. I think this is worth doing anyway. It has no effect on the user side, just tells Google that we're "official" and lets us point to site maps and such.

bryangingechen:
As Rob says, this verification is independent of Google Analytics. [Here's a page on the Google site verification](https://support.google.com/webmasters/answer/9008080?hl=en). It's mainly used in relation to the [Google search console](https://support.google.com/webmasters/answer/9128668?hl=en&ref_topic=9128571).

gebner:
Johan and I were commenting on Rob's proposal to add Google analytics (the javascript) to each page.  I am happy with this PR.

jcommelin:
Yup, I'm also fine with this PR.

PR #60, opened 2020-06-14
naming advice around `inj` and `ext`

semorrison:


templates/contribute/naming.md
```diff
@@ -259,6 +259,40 @@ check le_of_mul_le_mul_left
...
 ```
 
+## Naming of structural lemmas ##
+
+We are trying to standardize certain naming patterns for structural lemmas.
+At present these are not uniform across mathlib.
+
+### Extensionality ###
+
+A lemma of the form `(∀ a, a ∈ s₁ ↔ a ∈ s₂) → s₁ = s₂` should be named `.ext`,
```
urkud, line 269: We also have `∀ x, f x = g x` as a common assumption for `@[ext]`.

templates/contribute/naming.md
```diff
@@ -259,6 +259,40 @@ check le_of_mul_le_mul_left
...
+and labelled with the `@[ext]` attribute.
+Often this type of lemma can be generated automatically by putting the
+`@[ext]` attribute on a structure.
```
urkud, line 272: AFAIK `@[ext]` uses projections while quite often `ext` lemmas should use coercions.

semorrison: Good point. I will make this explicit.

semorrison, line 272:
```suggestion
Often this type of lemma can be generated automatically by putting the
`@[ext]` attribute on a structure.
(However an automatically generated lemma will always be written in terms
of the structure projections, and often there is a better statement,
e.g. using coercions, that should be written by hand then marked with `@[ext]`.)

```

templates/contribute/naming.md
semorrison:
```suggestion
A lemma of the form `(∀ x, f x = g x) → f = g` should be named `.ext`,
```

templates/contribute/naming.md
```diff
@@ -259,6 +259,40 @@ check le_of_mul_le_mul_left
...
+
+A lemma of the form `s₁ = s₂ ↔ ∀ a, a ∈ s₁ ↔ a ∈ s₂` should be named `.ext_iff`.
```
semorrison, line 274:
```suggestion
A lemma of the form `f = g ↔ ∀ x, f x = g x` should be named `.ext_iff`.
```

jcommelin:
I think it's good to codify what has been an implicit naming practice already. I'm fine with merging this. @bryangingechen @robertylewis @digama0 any comments?

bryangingechen:
I'm happy with merging this too.

robertylewis:
I'm not entirely up on the naming conventions but this looks fine to me. I don't think the trailing `###`s are necessary.

bryangingechen:
> I don't think the trailing `###`s are necessary.

They're just there for consistency with the rest of the file, I think. We can get rid of them in a cleanup pass later on. See #62 as well as the duplicated variable names table in style.md.

PR #63, opened 2020-06-15
overview.yaml: new links and some re-translations

hrmacbeth:
I added links to some newly-merged analysis facts ([#3059](https://github.com/leanprover-community/mathlib/pull/3059), [#3021](https://github.com/leanprover-community/mathlib/pull/3021), [#2660](https://github.com/leanprover-community/mathlib/pull/2660) ).  I also changed some translations that I found inelegant -- but I know that many of these may have been discussed already, so feel free to revert any.

### DIFF
--- a/data/overview.yaml
+++ b/data/overview.yaml
@@ -39,7 +39,7 @@ Linear algebra:
     special linear group: ''
     orientation of a $\R$-valued vector space: ''
   Matrices:
-    commutative ring valued matrices: 'data/matrix/basic.html#matrix'
+    commutative-ring-valued matrices: 'data/matrix/basic.html#matrix'
     field-valued matrices: 'data/matrix/basic.html#matrix'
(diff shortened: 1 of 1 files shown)


PatrickMassot:
I guess you learned all those theorems in an English-speaking country, so this is probably better. The only thing I find more difficult to understand is this Taylor-Peano thing. Maybe we could write "Taylor theorem with little-o remainder"?

hrmacbeth:
Yes, I like that, it is very clear.  Actually I have not seen this "Peano" version in the courses I have taken or taught; they had only the integral form and the Lagrange form.  It is interesting to see how the syllabus varies.

PatrickMassot:
Very interesting indeed. In France the little-o remainder is the basic version, and everything else is presented as a refinement. I think it comes from the fundamental link with the definition of differentiability.

PR #67, opened 2020-06-19
simpler links and ci for overview

robertylewis:
This does something similar to #54 for the undergrad overview page. Not printing types, but computing links from decl names using the doc-gen database.

I found and fixed a few broken links with this. There's another that I wasn't immediately sure how to fix. Some of the links are to pages rather than specific declarations, what should we do here?


PatrickMassot:
Originally we discussed not to do this because it's easier for authors of this yaml file to search for stuff on the API docs website and then copy paste part of the url instead of fighting to find the full name of something. There is also the issue that sometimes the most relevant link is a full file. We still need CI to check all links, and this is not done currently.

What is the advantage of doing what you did?

bryangingechen:
It's true that we're not checking for all broken links, but that's work for a separate PR, I think.

While it shouldn't be too hard to check for 404's, most links to mathlib_docs won't just point to a page, but to a specific anchor link. I think it would be a pain to write a CI script that could check that internal anchor links were present, right? 

If I understand correctly, contributors need only copy paste the declaration name from the mathlib docs, which isn't too much harder than copying part of the URL. (We could even change doc-gen so that there's a little icon next to the declaration names that puts the name in the clipboard when cllcked.)

PatrickMassot:
You're right, I forgot that the docs displays the full name.

robertylewis:
This change also checks links. When you run `make_site.py`, it will print each declaration that it didn't find in the docs database. Change the print to an exception and the site won't build unless all the links work.

For the links to full pages, we could e.g. use slightly different syntax to enter them in the yaml file and check for that.

robertylewis:
Another upside of this is that it's more robust to refactoring. If a declaration is moved to another file (but keeps its name) the link won't break.

PatrickMassot:
Ok, I'm fine with this change. And I think we should indeed be warned when the build is not successful.

robertylewis:
This wasn't ready to merge! 

> For the links to full pages, we could e.g. use slightly different syntax to enter them in the yaml file and check for that.

isn't implemented so links to full pages are broken.

I'm going to revert the commit and update this later, I don't have time at the moment.

bryangingechen:
Ah, sorry about that!

PR #68, opened 2020-06-23
simpler links and ci for overview

robertylewis:
The return of #67. This should properly handle links to files, and will throw an error on a bad link instead of just printing it. Also fixes a few bad links.


PatrickMassot:
Can I assume you tested this locally?

robertylewis:
Yes, you can!

A warning: once this is merged, we'll start seeing intermittent failures on CI for this repo, as mathlib changes cause the overview links to break. This could be confusing because the errors will show up in PRs/daily builds here, despite no relevant changes in this repo.

bryangingechen:
LGTM. I think the errors that pop up should be easy enough to track down. Though we should probably be stricter about asking folks to include renames in the commit message bodies (except possibly in cases where there are tons of declarations renamed with an obvious pattern).

PR #69, opened 2020-06-25
fix(data/100): correct decl for item 20

jsm28:
This entry points to the wrong theorem called `sum_two_squares`.


bryangingechen:
Thanks!

PR #70, opened 2020-06-26
Update contribute/index.md

paulvanwamelen:


jcommelin:
Thanks! Looks good to me :+1:

bryangingechen:
This seems fine to me too, though I hope we aren't planning to be too strict on requiring people to PR only from mathlib branches.

PatrickMassot:
I don't mind putting pressure on this. It will also help enforcing the instruction to discussing things on Zulip before too much.

PR #71, opened 2020-06-26
doc(Update project.md)

paulvanwamelen:
Add detailed steps/instructions for contributing to mathlib. (Copied from Patrick's Zulip messgae)


PatrickMassot:
Thanks, but this is the wrong file.  Those things should go to https://github.com/leanprover-community/leanprover-community.github.io/blob/newsite/templates/contribute/index.md. You can still add a link between those files if you expected to find the information here.

PR #74, opened 2020-06-29
propose to use "pip install ... --user", etc

dimpase:


bryangingechen:
Thanks! Do you mind copying this text to the instructions to `linux.md`, `macos.md` and `windows.md`?

PatrickMassot:
We're not interested in merging insults to the community.

dimpase:
who am I inslting here? I am trying to help.

jcommelin:
@PatrickMassot Could you clarify a bit? I personally don't think `sudo whatever` is very dangerous, but as far as I can see this PR doesn't do anything weird.

b-mehta:
@jcommelin On the note of `sudo pip install`, it's considered bad practice and to be avoided where possible, see for instance [this ubuntu question](https://askubuntu.com/questions/802544/is-sudo-pip-install-still-a-broken-practice) and [this SO question](https://stackoverflow.com/questions/33004708/osx-el-capitan-sudo-pip-install-oserror-errno-1-operation-not-permitted/33004920#33004920), which includes

"Instructions telling sudo pip install are inherently wrong.

If there is any tutorial out there which says you should do sudo pip then please file a bug against this package. The author is dis-educating Python community, as time has proven sudo pip to be a broken practice."

jcommelin:
@b-mehta Aha, I didn't know this...

PR #77, opened 2020-06-30
Use pipx for installation of mathlibtools

TwoFX:
This is my attempt at the most likely abandoned #74, trying to address the fact that using `sudo pip3 install` seems to be discouraged (cf. [this comment](https://github.com/leanprover-community/leanprover-community.github.io/pull/74#issuecomment-651815192)). The easiest way of installing end-user software from PyPI that follows "best practices" that I am aware of is `pipx`, which in particular has a command `ensurepath` that does all the `PATH` fiddling (which is not necessary when using `sudo`, but is necessary when using `--user`) automatically. This PR changes the instructions to follow this route.

### DIFF
--- a/templates/install/windows.md
+++ b/templates/install/windows.md
@@ -44,9 +44,12 @@ In order to use mathlib supporting tools, you need to [get python](https://www.p
 * Open Git Bash (type `git bash` in the Start Menu)
 * Run `which python`
   * The expected output is something like `/c/Users/<user>/AppData/Local/Programs/Python/Pythonxx-xx/python`. In this case, proceed to the next step.
-  * If it's something like `/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python`, then you need to disable a Windows setting.
-    * Type `manage app execution aliases` into the Windows search prompt (start menu) and open the corresponding System Settings page.
-    * There should be two entries `App Installer python.exe` and `App Installer python3.exe`. Ensure that both of these are set to `Off`.
+  * If it's something like `/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python`, then
+    * Did you follow the instruction to select `Add Python 3.x to PATH` during the installation?
+      * If not, re-run the python installer to uninstall python and try again.
+    * Otherwise, you need to disable a Windows setting.
+      * Type `manage app execution aliases` into the Windows search prompt (start menu) and open the corresponding System Settings page.
+      * There should be two entries `App Installer python.exe` and `App Installer python3.exe`. Ensure that both of these are set to `Off`.
     * Close and reopen Git Bash and restart this step.
   * If it is any other directory, you might have an existing version of Python. Ask for help in the Zulip chat room (linked above).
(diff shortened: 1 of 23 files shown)


dimpase:
I'd have mentioned what `pipx` does, e.g.
```
(along with a support tool [pipx](https://pipxproject.github.io/pipx/))
```

TwoFX:
> I'd have mentioned what `pipx` does, e.g.
> 
> ```
> (along with a support tool [pipx](https://pipxproject.github.io/pipx/))
> ```

Good point. I've added this and also took the liberty to slightly homogenize the different guides.

templates/install/windows.md
```diff
@@ -62,10 +62,15 @@ In order to use mathlib supporting tools, you need to [get python](https://www.p
...
-  pip3 install mathlibtools
-  ```
+Then we install a small tool called `leanproject` that which will handle
+updating mathlib according to the needs of your current project. We use
+[pipx](https://pipxproject.github.io/pipx/) to perform the installation.
+At a terminal, run the commands
+```bash
+python3 -m pip install --user pipx
+python3 -m pipx ensurepath
+pipx install mathlibtools
```
semorrison, line 72: Sadly, this doesn't work. I just followed the instructions on a fresh Windows 10 machine, and upon getting to this step `pipx install mathlibtools` just says `bash: pipx: command not found`.

I closed and restarted "git bash", and get the same message. I rebooted the windows machine (I have not yet tried unplugging the router...) and get the same message.

TwoFX: Thank you for testing this. I'm looking into it.

TwoFX:
Despite `pipx` documentation claiming that it works on Windows, I was unable to confirm that this is indeed the case, so I decided to roll back the changes. Global `pip install` doesn't require administrator privileges on Windows anyway, the difference between global and user install is merely whether the package is installed to `AppData/Local` (global) or `AppData/Roaming` (User), so personally I think manually having the user figure out the correct path (which includes both the user name and the Python version number) is not worth it. I have also rolled back the changes to `macos.md`; while i believe that it should work similar to Linux (replacing `apt` with `brew`), I have no way to verify this.

templates/install/windows.md
TwoFX: The first problem is that `pipx ensurepath` modifies the Windows `PATH` variable, but we want to modify the `git bash` `PATH`, which are different.

After manually setting the path, there was also some `venv` problem, which can probably be fixed, but I didn't try very hard.

PatrickMassot:
@TwoFX am I correct in thinking that instructions in this PR now match what is done by the automatic scripts, hence work? Could you have a look at the git conflict?

TwoFX:
> @TwoFX am I correct in thinking that instructions in this PR now match what is done by the automatic scripts, hence work? 

Yes the instructions match.

> Could you have a look at the git conflict?

I merged; is the conflict resolved?

PatrickMassot:
Actually, the conflicts were due to the complicated git history. The end result was not conflicting, so I squashed. Thanks again for looking into all this.

PR #78, opened 2020-07-01
Gives titles to pages

### LINKED ISSUE #75: HTML `<title>`s
Right now, the `<title>`s of all the pages just say `Lean community`, which isn't brilliant when a user may have multiple pages open at the same time, or if they want to search for something, as they'll have to look at the URL to find what they want.

Also it might help when people are looking for things using a search engine? Although right now Google is the only search engine that even finds any of the install pages, and it's smart enough to pull in the `<h1>`.

It might make more sense to use the markdown titles for the HTML `<title>` tags.

PatrickMassot:
Fixes #75

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -11,7 +11,7 @@ import yaml
 from staticjinja import Site
 import jinja2.ext
 import pybtex.database
-from mistletoe import Document, HTMLRenderer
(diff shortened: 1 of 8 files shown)


bryangingechen:
LGTM. I'll fix the titles for the "100" pages later.

PR #79, opened 2020-07-08
reminder to run git commands in the mathlib repo

bryangingechen:
See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/equivalent.20of.20squeeze_simp.20for.20linarith.2C.20group.2C.20etc.3F/near/203230223).


bryangingechen:
See #80.

PR #88, opened 2020-07-11
add data structures to overview

robertylewis:


data/overview.yaml
```diff
@@ -261,6 +261,34 @@ Analysis:
...
+    array: 'array'
+    buffer: 'buffer'
+    difference list: 'dlist'
+    lazy list: 'lazy_list'
+    stream: 'stream'
+    sequence: 'seq'
+    weak sequence: 'wseq'
+    
+  Sets:
+    finite set: 'finset'
```
jcommelin, line 276:
```suggestion
    set: 'set'
    finite set: 'finset'
```

PatrickMassot:
It feels weird to see sets and finsets in data structures which sounds like pure computer science. To be honest I didn't know where to put them in the other categories.

robertylewis:
`finset` has computational content and could certainly be used for "computer science." `set` doesn't, which is why I didn't put it on the list. We could call the section "combinatorial structures" or something like that if you prefer, I don't care.

jcommelin:
I think we have to include `set` somewhere... I don't have much opinion about where exactly.

PatrickMassot:
Maybe we should create a basics section. It could confuse people who think sets are the foundations of mathematics, but probably we won't have so many of those.

jcommelin:
Ok, that is a good option. :+1:

robertylewis:
What do you imagine in this "basics" section? Should it just be the "datatypes" section in this PR with `set` added?

robertylewis:
I'm not so sure about my inclusion of the meta rb_maps/sets in this PR, btw. This is really a list of things we have formal theories of and we can't prove things about those. So I think I should remove them, but I'll wait until we decide on other changes.

jcommelin:
I'm also happy with just leaving `set` out. It shouldn't hold up this PR for so long.

robertylewis:
Okay, for now aesthetics win out. I deleted the meta rb-stuff, so to keep at least three entries in the `sets` section, I added `set`. I'm fine merging this as is, or renaming, or reogranizing, but this is my best idea at the moment.

robertylewis:
Is there a consensus about a better way to arrange this? I'm okay with the current state of this PR and don't have a better plan.

PR #96, opened 2020-07-18
Add tl;dr

kbuzzard:
This page is far too long IMNSHO.

### DIFF
--- a/templates/mwe.md
+++ b/templates/mwe.md
@@ -1,5 +1,31 @@
 # Minimal working examples
 
+## tl;dr
+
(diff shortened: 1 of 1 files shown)


templates/mwe.md
```diff
@@ -1,5 +1,31 @@
...
+variable (X : Type u)
+
+open set
+
+#check (univ : set X)
+```
+
+## Formal Definition
+
 A **minimum working example** is a code snippet that can be copied-and-pasted into an empty Lean file and still have the same features (working) and that does not include unnecessary details (minimum).
```
robertylewis, line 29:
```suggestion
A **minimal working example** is a code snippet that can be copied-and-pasted into an empty Lean file and still have the same features (working) and that does not include unnecessary details (minimal).
```

(not related to your changes, but to match the title above)

PR #97, opened 2020-07-18
display decl counts on stats page

robertylewis:
image

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -142,6 +142,10 @@ with gzip.GzipFile('export_db.json.gz', 'r') as json_file:
 
 decl_loc_map = json.loads(json_bytes.decode('utf-8'), strict=False)
 
+num_thms = len([d for d in decl_loc_map if decl_loc_map[d]['kind'] == 'thm'])
+num_meta = len([d for d in decl_loc_map if decl_loc_map[d]['is_meta']])
+num_defns = len(decl_loc_map) - num_thms - num_meta
+
 def undecorate_arg(arg):
     return ''.join(
         match[4] if match[0] == '' else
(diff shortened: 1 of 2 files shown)


robertylewis:
(CI should pass after the next docs build.)

bryangingechen:
Do we want to make graphs of these too?

robertylewis:
> Do we want to make graphs of these too?

This is much harder than the other graphs since each data point takes a compiled mathlib to calculate. It's far too expensive to regenerate the whole data set each time, like gitstats does for the others. We'd need to add points incrementally and store a cache somewhere. And seeding the first one will take a huge amount of time. Let's say 2.5 years before we started storing olean caches, one data point a day instead of per commit like gitstats, 15 min per build average (older ones were faster). That's 10 days of computing -- better parallelize that!

bryangingechen:
Ah, OK. I'm not sure we have to have data points for all the old commits for a graph to be useful, but there are some other interesting things we might want to compute from each mathlib commit (like the type class graph and maybe some statistics on that), so maybe we can look into a unified way of caching the history of those later on.

PatrickMassot:
I'm on vacation this week. Both my computer at home and my office computer are idle (and fast). If you give me scripts to run I'd be happy to help.

PatrickMassot:
And of course we could restrict to one data point per week in the archive.

robertylewis:
The script is nontrivial and there's the issue of where to store the cached results, which is going to force some annoying Actions wrangling. I can't say I'm excited about doing this just to produce a graph which, I predict, will be a slightly flatter version of the LOC graph. If someone wants to do this I certainly won't stop them but I can't justify the effort myself.

PR #101, opened 2020-07-25
Using nat instead of ℕ for natural numbers

ikrukov:
Toolchain.md contains introductory materials. Using unicode symbol for natural numbers may confuse newcomers - it can be not so clear how to enter ℕ symbol.


PatrickMassot:
Thank you very much for trying to improve our documentation! However I very strongly disagree with this edit and your justification. This file is not meant for beginners, except maybe those who actively try to complicate things by refusing to use our tooling. The word nat itself is infinitely harder to read than the symbol, and writing such symbols is the first thing you need to learn when staring your Lean training anyway.

PR #105, opened 2020-08-06
Mention tactic `extract_goal` in `mwe.md`

utensil:
See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Being.20stuck.20-.3E.20MWE/near/206124531).

When people are asked to provide a `#mwe` on Zulip, they can directly see the tip about using `extract_goal` to create a `#mwe`.

### DIFF
--- a/templates/mwe.md
+++ b/templates/mwe.md
@@ -24,7 +24,8 @@ open set
 #check (univ : set X)
 ```
 
-Tip: If you are using [Mathlib](https://github.com/leanprover-community/mathlib) and have `import tactic` in your file, there's a tactic called [extract_goal](https://leanprover-community.github.io/mathlib_docs/tactics.html#extract_goal) that can help you format the current goal as a stand-alone example, you just need to also include the corresponding `import`s and `open`s and `universe`s and `variable`s to create a minimal working examples.
+Tip: If you are using [Mathlib](https://github.com/leanprover-community/mathlib) and have `import tactic` in your file, there's a tactic called [`extract_goal`](https://leanprover-community.github.io/mathlib_docs/tactics.html#extract_goal) that can help you format the current goal as a stand-alone example.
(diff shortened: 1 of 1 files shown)


templates/mwe.md
```diff
@@ -24,6 +24,8 @@ open set
 #check (univ : set X)
 ```
 
+Tip: If you are using [Mathlib](https://github.com/leanprover-community/mathlib) and have `import tactic` in your file, there's a tactic called [extract_goal](https://leanprover-community.github.io/mathlib_docs/tactics.html#extract_goal) that can help you format the current goal as a stand-alone example, you just need to also include the corresponding `import`s and `open`s and `universe`s and `variable`s to create a minimal working examples.
```
eric-wieser, line 27:
```suggestion
Tip: If you are using [Mathlib](https://github.com/leanprover-community/mathlib) and have `import tactic` in your file, there's a tactic called [`extract_goal`](https://leanprover-community.github.io/mathlib_docs/tactics.html#extract_goal) that can help you format the current goal as a stand-alone example.
Note you still need to include the corresponding `import`s and `open`s and `universe`s and `variable`s as mentioned above.
```

PatrickMassot:
I would also suggest to remove extra objects and assumptions from the output of `extract_goal`. But maybe that would end up being a complicated advice compared to the goal of producing a mwe.

utensil:
> I would also suggest to remove extra objects and assumptions from the output of `extract_goal`.

How to determine whether an object or an assumption is redundant?

> But maybe that would end up being a complicated advice compared to the goal of producing a mwe.

This can be solved by providing an option?

bryangingechen:
I think Patrick is simply proposing to add text saying something like this: "Remember that you can remove extraneous variables and hypotheses from the output of `extract_goal` to further minimize your example."

PR #108, opened 2020-08-10
feat(data/100): Formula for Pythagorean Triples

jsm28:
This has been in mathlib for a month, but didn't get added to this
list when it was added to mathlib.


bryangingechen:
Thanks!

PR #111, opened 2020-08-14
New export_db.json format from docs.

gebner:
Instead of parsing the U+E000 syntax here, we let the documentation generator render an HTML version.  Now all the names, notation, etc. are hyperlinked.  I changed the link color to black because the website blue is jarring when used this much.

Depends on https://github.com/leanprover-community/doc-gen/pull/49

### DIFF
--- a/scss/lean.scss
+++ b/scss/lean.scss
@@ -46,20 +46,37 @@ li.nav-item.dropdown:hover .dropdown-menu {
   background-color: white !important;
 }
 
+/* Theorem statements in the 100 theorem list: */
+
 .doc-stmt {
-  font-family: monospace;
+    font-family: monospace;
+    /* indent everything but first line twice as much as decl_type */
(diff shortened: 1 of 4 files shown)


gebner:
Corresponding website PR: https://github.com/leanprover-community/doc-gen/pull/49

scss/lean.scss
```diff
@@ -46,20 +46,37 @@ li.nav-item.dropdown:hover .dropdown-menu {
...
 
-.doc-stmt.nm {
-  font-weight: bold;
+.doc-stmt .decl_type {
+    margin-top: 2px;
+    margin-left: 4ex; /* extra indentation */
+}
+
+.doc-stmt .impl_arg {
+    font-style: italic;
```
robertylewis, line 63:
```suggestion
```
I think this looks pretty odd in the example you posted.

PR #113, opened 2020-08-16
Update undergrad.yaml

hrmacbeth:


data/undergrad.yaml
```diff
@@ -294,14 +294,14 @@ Single Variable Real Analysis:
...
     additive subgroups of $\R$:
   Numerical Series:
-    Convergence of real valued-series:
+    Convergence of real valued-series: 'has_sum'
     Geometric series: 'has_sum_geometric_of_abs_lt_1'
-    Riemann series:
-    positive valued series:
-    summation of comparison relations:
+    Riemann series (i.e., individual values of the Riemann zeta function):
+#    positive valued series: (this is certainly a concept that can be expressed using the definitions in mathlib, but there is no particular theorem)
```
PatrickMassot, line 300: No, we cannot express this yet since we don't even have series. This item is a bit vague but it refers to a lot of results that hold only for series with positive values (for instance if such a series has general term bounded above by the term of a convergent series then it converges).

data/undergrad.yaml
```diff
@@ -327,22 +327,22 @@ Single Variable Real Analysis:
...
-    reciprocal hyperbolic trigonometric functions:
+    inverse (circular) trigonometric functions: 'real.arcsin'
+    inverse hyperbolic trigonometric functions:
   Integration:
     integral over a segment of piecewise continuous functions:
     antiderivatives:
     Riemann sums:
     antiderivative of a continuous function:
-    usual methods to compute integrals:
+    the usual methods to compute integrals:
```
PatrickMassot, line 345: Actually I think this was meant as a subheading for the next two items so we should remove that line.

data/undergrad.yaml
```diff
@@ -294,14 +294,14 @@ Single Variable Real Analysis:
     connected subsets of $\R$: 'set_of_is_preconnected_eq_of_ordered'
     additive subgroups of $\R$:
   Numerical Series:
-    Convergence of real valued-series:
+    Convergence of real valued-series: 'has_sum'
     Geometric series: 'has_sum_geometric_of_abs_lt_1'
-    Riemann series:
-    positive valued series:
-    summation of comparison relations:
+    Riemann series (i.e., individual values of the Riemann zeta function):
```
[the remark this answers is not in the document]
hrmacbeth: I'm confused, are these not the same? https://en.wikipedia.org/wiki/Riemann_zeta_function

data/undergrad.yaml
hrmacbeth: We have `summable_of_norm_bounded` for the theorem you mentioned, but I guess it's a question of how literally to take the list.  Anyway, just let me know what you'd like to do -- should I clear the whole series section or only the ones you marked?  Feel free to make the changes yourself, also.

data/undergrad.yaml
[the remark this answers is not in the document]
PatrickMassot: The emphasis is not at all on values of the sum, it's only about convergence.

data/undergrad.yaml
PatrickMassot: I mean clearing the whole series section. It will be easy to fill once someones starts putting the relevant definitions and API, but we have none of that yet. This is not so important. I guess this whole section is very French, and for a very stupid reason: convergence of numerical series is great to create tricky technical exercises, and France has a lot of competitive exams at that level.

hrmacbeth: Ok, I cleared it all, including the pre-existing entry for geometric series (but I can put that one back if you prefer).

data/undergrad.yaml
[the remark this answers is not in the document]
hrmacbeth: I have often seen it called the `p`-series in US textbooks, so let's go with that.

PatrickMassot:
Thanks!

PR #118, opened 2020-08-20
Update 100.yaml

jjaassoonn:
add "e is transcendental"


jcommelin:
You also did "pi is transcendental", right?

robertylewis:
There seem to be disproportionately many results about transcendental numbers on this list...

jcommelin:
Or Euclidean geometry for that matter.

jjaassoonn:
> You also did "pi is transcendental", right?
> 
> But probably not Hermite–Lindemann?

No, I didn't do $pi$ because it relies on symmetric polynomial

PR #135, opened 2020-10-01
Allow semicolons in single line tactic proofs.

gebner:
Since version 3.13, Lean shows the goal state after tactics separated by `;`.  As far as I remember, the main reason to avoid `;` was that the proofs were hard to inspect.  Since this is now fixed in Lean, we should allow semicolons again.


gebner:
According to the poll on Zulip, people still don't like semicolons.

https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.3B.20in.20by

PR #137, opened 2020-10-01
add favicon

robertylewis:
The docs and Zulip archive both use the "L" icon (copied shamelessly from the MS Lean page). Should we use that on the main site too?


gebner:
Definitely!

PR #138, opened 2020-10-01
Show complete menu on the left.

gebner:
I find it very hard to find certain pages on the website, such as the style guidelines, the theorem lists, or really anything that doesn't have a prominent link on the first page.  If I didn't know that we had these pages, I would never find them.

This PR proposes to add reasonably complete list of all pages in a navigation bar on the left.

I didn't polish this too much, neither the design nor the organization of the pages under the headings, because I wanted to get some feedback first.

navpanel


robertylewis:
We're thinking alike! I posted something earlier today on Zulip about reorganizing the website and was going to try to find time tomorrow. Thanks for saving me the trouble :) 

I think this setup looks awkward in a full screen window. There's a sea of white between the menu and content. Also the giant empty gray bar at the top doesn't add anything.

Not sure what the right design fix is. One possibility is keeping the current drop down style but using your organization. Maybe this is too many menus but I think it could work.

gebner:
I am strongly against dropdowns, they hide the content and you can't search them using the browser's search function.

> I think this setup looks awkward in a full screen window. There's a sea of white between the menu and content.

The content was a bit off-center, I've fixed this now in case this irritated you as well.  Do you like the style used in the API docs, where the navigation panel is directly next to the content?  I kind of like that the navigation bar is a bit out of the way here.

robertylewis:
Ah, I guess the content was centered in the whitespace before instead of in the page? This looks better.
On the docs pages, I think the content has to be as wide as possible (we're displaying code), but here a max width on the content seems appropriate. Maybe the max width could be wider.

There's some slightly odd behavior as you shrink the window, at least in Chrome (85 this time): there's a range where it balances the left sidebar with whitespace on the right, shrinking the content; then a range where it gives up on the right whitespace; then a range where it shows both the sidebar and the dropdown icon, but the icon doesn't do anything; then eventually the sidebar goes away and the dropdown works.

gebner:
> On the docs pages, I think the content has to be as wide as possible

We're also setting a max width on the API docs.  It becomes extremely hard to read when the text becomes very wide.  I've increased the max width to make it the same as the API docs.

> There's some slightly odd behavior as you shrink the window, at least in Chrome

Are you on the latest version (of the website)?  There is the intermediate step where it gives up the space on the right, but then it switches directly from gray bar at the left to gray bar at the top.

robertylewis:
> We're also setting a max width on the API docs.

Oh -- it doesn't reach the max width on my 1920-sidebar px wide laptop, which is wide enough for my purposes! Comparing to your latest commit, it looks like it's just on the edge.

I like this width much more, and yeah, looks like you fixed the weird intermediate stage!

My last complaint: the logo on the home page is huge. Let's scale that down. Then I think I love the layout.

templates/_base.html
```diff
@@ -36,40 +34,33 @@
...
-    </div>
   </nav>
-  <div class="container-sm mt-4">
+  <div class="container-fluid"><div class="row">
+    <nav id="navbar" class="col-md-3 col-xl-2 d-md-block bg-light sidebar collapse navbar navbar-light pb-4">
+      <div class="d-sm-none d-md-block"><a class="navbar-brand" href="{{ base_url }}index.html">Lean Community</a></div>
+      {% for menu in menus %}<div>
+        <h6 class="sidebar-heading px-3 mt-4 mb-1">{{ menu.title }}</h6>
+        <ul class="nav flex-column">{% for item in menu.items %}
+          <li class="nav-item"><a href="{{ item.url | url }}">{{ item.title }}</a></li>
```
eric-wieser, line 53:
```suggestion
          <li class="nav-item"><a class="nav-link" href="{{ item.url | url }}">{{ item.title }}</a></li>
```

templates/_base.html
```diff
@@ -36,40 +34,33 @@
...
-          </div>
-        </li>
-		{% endfor %}
-      </ul>
-    </div>
   </nav>
-  <div class="container-sm mt-4">
+  <div class="container-fluid"><div class="row">
+    <nav id="navbar" class="col-md-3 col-xl-2 d-md-block bg-light sidebar collapse navbar navbar-light pb-4">
+      <div class="d-none d-md-block"><a class="navbar-brand" href="{{ base_url }}index.html">Lean Community</a></div>
```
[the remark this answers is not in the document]
eric-wieser: Nevermind, only @robertylewis's hosted version has a problem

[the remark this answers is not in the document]
gebner: I've already fixed it right after your Zulip comment.

templates/_base.html
gebner: Ok, I've used `d-block` though.  The `nav-link` class messes with the padding.

templates/_base.html
```diff
@@ -36,40 +34,33 @@
...
-  <div class="container-sm mt-4">
+  <div class="container-fluid"><div class="row">
+    <nav id="navbar" class="col-md-3 col-xl-2 d-md-block bg-light sidebar collapse navbar-light pb-4">
+      <div class="d-none d-md-block"><a class="navbar-brand" href="{{ base_url }}index.html">Lean Community</a></div>
+      {% for menu in menus %}<div>
+        <h6 class="sidebar-heading px-3 mt-4 mb-1">{{ menu.title }}</h6>
+        <ul class="nav flex-column">{% for item in menu.items %}
+          <li class="nav-item"><a class="d-block" href="{{ item.url | url }}">{{ item.title }}</a></li>
+        {% endfor %}</ul>
+      </div>{% endfor %}
```
eric-wieser, line 55: IMO this leads to better layout on mobile:
```suggestion
      <div class="row">
      {% for menu in menus %}<div class="col-12 col-sm-6 col-md-12">
        <h6 class="sidebar-heading px-3 mt-4 mb-1">{{ menu.title }}</h6>
        <ul class="nav flex-column">{% for item in menu.items %}
          <li class="nav-item"><a class="d-block" href="{{ item.url | url }}">{{ item.title }}</a></li>
        {% endfor %}</ul>
      </div>{% endfor %}
      </div>
```

gebner: I don't really have a strong opinion here, so let's do the two-column layout.

robertylewis:
One more nitpick. The margin at the top of the content is a lot bigger than on the sidebar. I think it looks unbalanced:

image


Maybe the sidebar should move down slightly and the content should move up?

gebner:
I moved the "Lean Community" header down a bit.

robertylewis:
Another quirk, at the bottom of https://robertylewis.com/mathlib/ci.html

image

PR #139, opened 2020-10-01
Tweak styles to be similar to API docs.

gebner:
Depends on #138.

Adds a few more style tweaks to make the website more similar to the API docs:
- Link color is darkened a bit.
- Open Sans as font family for regular text.
- Source Code Pro for code blocks.
- Code and code blocks have a slightly gray background, the red-purple color is removed.


robertylewis:
It's beautiful! I appreciate the consistency with the API docs.

Maybe the nav panel needs to be slightly wider in this style? Or the "Lean Community" title slightly smaller. The text is just a bit wider than the max width and forces a line break.

gebner:
> Maybe the nav panel needs to be slightly wider in this style? Or the "Lean Community" title slightly smaller. The text is just a bit wider than the max width and forces a line break.

The line break was actually forced so that we can have a large font size. :smile:  I've reduced the font size now so that it doesn't break;

gebner:
Merged into #138

PR #140, opened 2020-10-02
reorganizing menu items

robertylewis:
Whether or not these are displayed as a sidebar or top menus, here are some suggestions for ordering and organizing! Very open to comments.

This is a PR into the `navpanel` branch.

### DIFF
--- a/data/menus.yaml
+++ b/data/menus.yaml
@@ -1,15 +1,24 @@
+- title: Community
+  items:
+    - Zulip chat: https://leanprover.zulipchat.com/
+    - GitHub: https://github.com/leanprover-community/mathlib
+    - Community information: meet.html
+    - Papers about Lean: papers.html
+    - Projects using Lean: lean_projects.html
+
 - title: Installation
   items:
     - Get started: get_started.html
-    - Linux installation: install/linux.html
-    - Debian installation: install/debian.html
+    - Debian/Ubuntu installation: install/debian.html
+    - Generic Linux installation: install/linux.html
     - MacOS installation: install/macos.html
     - Windows installation: install/windows.html
     - Online version (no installation): https://leanprover-community.github.io/lean-web-editor
 
 - title: Documentation
   items:
-    - Big list of docs (start here): learn.html
+    - Leaning resources (start here): learn.html
+    - API documentation: https://leanprover-community.github.io/mathlib_docs
     - calc mode: extras/calc.html
     - conv mode: extras/conv.html
     - simplifier: extras/simp.html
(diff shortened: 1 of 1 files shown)


data/menus.yaml
```diff
@@ -1,15 +1,24 @@
+- title: Community
```
robertylewis, line 1: I think the Zulip and GitHub links need to be at the top. The others are much less important but they make sense in this section.

data/menus.yaml
```diff
@@ -1,15 +1,24 @@
...
+    - Community information: meet.html
+    - Papers about Lean: papers.html
+    - Projects using Lean: lean_projects.html
+
 - title: Installation
   items:
     - Get started: get_started.html
-    - Linux installation: install/linux.html
-    - Debian installation: install/debian.html
+    - Debian/Ubuntu installation: install/debian.html
```
robertylewis, line 12: My prediction: a lot of Ubuntu users won't recognize the name Debian.

data/menus.yaml
```diff
@@ -18,14 +27,12 @@
...
 
 - title: Library overviews
   items:
     - Library overview: mathlib-overview.html
-    - Wiedijk's 100 theorems: 100.html
     - Undergraduate maths: undergrad.html
-    - Other Lean projects: lean_projects.html
+    - Wiedijk's 100 theorems: 100.html
 
 - title: Theory docs
```
robertylewis, line 37: I'm not convinced this needs to be on the side panel. These are super out of date. (The p-adics one is my fault.) In fact, I think there's an argument for removing them from the site entirely, until we have a procedure for making sure they stay updated, but that's another discussion.

jcommelin:
lgtm

data/menus.yaml
```diff
@@ -1,15 +1,24 @@
...
+    - Debian/Ubuntu installation: install/debian.html
+    - Generic Linux installation: install/linux.html
     - MacOS installation: install/macos.html
     - Windows installation: install/windows.html
     - Online version (no installation): https://leanprover-community.github.io/lean-web-editor
 
 - title: Documentation
   items:
-    - Big list of docs (start here): learn.html
+    - Leaning resources (start here): learn.html
```
gebner, line 20:
```suggestion
    - Learning resources (start here): learn.html
```

Unless this is supposed to be a clever pun.

data/menus.yaml
gebner: The side panel should be a complete list of what's on the web site; otherwise people get lost and can't find the page they're looking for.

My secret hope is that making this pages more visible will motivate other people to clean them up (or maybe turn them into module doc strings).

gebner (approved):

data/menus.yaml
robertylewis: Hmm. I'm just going to delete the p-adics one now. It predates module docs and is subsumed already. Looking at them all again, I can see arguments for keeping especially the natural numbers, sets, and topology writeups. LA and CT seem outdated.

PR #143, opened 2020-10-05
Clarify line breaking and indentation for theorems/definitions

Vierkantor:
Some clarifications in the style guide.

 * These rules apply for `def` and for `lemma`/`theorem`
 * Indent assumptions / conclusion after a line break
 * Short theorems can also fit on one line.

### DIFF
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -92,7 +92,7 @@ New bibliography entries should be added to `docs/references.bib`.
 
 See our [documentation requirements](doc.html) for more suggestions and examples.
 
-### Structuring definitions and theorems ###
+### Structuring definitions and theorems
 
 These guidelines hold for declarations starting with `def`, `lemma` and `theorem`. For "theorem statement", also read "type declaration of a definition" and for "proof" also read "definition body".
 
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -94,7 +94,7 @@ See our [documentation requirements](doc.html) for more suggestions and examples
 
 ### Structuring definitions and theorems
 
-These guidelines hold for declarations starting with `def`, `lemma` and `theorem`. For "theorem statement", also read "type declaration of a definition" and for "proof" also read "definition body".
+These guidelines hold for declarations starting with `def`, `lemma` and `theorem`. For "theorem statement", also read "type of a definition" and for "proof" also read "definition body".
 
 Use spaces around ":", ":=" or infix operators. Put them before a line break rather
 than at the beginning of the next line.
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -94,6 +94,8 @@ See our [documentation requirements](doc.html) for more suggestions and examples
 
 ### Structuring definitions and theorems ###
 
+These guidelines hold for declarations starting with `def`, `lemma` and `theorem`. For "theorem statement", also read "type declaration of a definition" and for "proof" also read "definition body".
+
 Use spaces around ":", ":=" or infix operators. Put them before a line break rather
 than at the beginning of the next line.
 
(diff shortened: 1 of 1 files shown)


templates/contribute/style.md
```diff
@@ -94,6 +94,8 @@ See our [documentation requirements](doc.html) for more suggestions and examples
 
 ### Structuring definitions and theorems ###
 
+These guidelines hold for declarations starting with `def`, `lemma` and `theorem`. For "theorem statement", also read "type declaration of a definition" and for "proof" also read "definition body".
```
Vierkantor, line 97: Anyone know a good way to express "type declaration of a definition"?

templates/contribute/style.md
```diff
@@ -107,6 +109,22 @@ theorem nat_case {P : nat → Prop} (n : nat) (H1: P 0) (H2 : ∀m, P (succ m))
...
+namespace nat
+
+lemma le_induction {P : ℕ → Prop} {m}
+  (h0 : P m) (h1 : ∀ n, m ≤ n → P n → P (n + 1)) :
+  ∀ n, m ≤ n → P n :=
+by apply nat.less_than_or_equal.rec h0; exact h1
+
+def decreasing_induction {P : ℕ → Sort*} (h : ∀ n, P (n + 1) → P n) {m n : ℕ} (mn : m ≤ n)
+  (hP : P n) : P m :=
+le_rec_on mn (λ k ih hsk, ih $ h k hsk) (λ h, h) hP
```
eric-wieser, line 123: Playing devils advocate - is this more readable as
```suggestion
lemma le_induction {P : ℕ → Prop} {m}
(h0 : P m) (h1 : ∀ n, m ≤ n → P n → P (n + 1)) :
  ∀ n, m ≤ n → P n :=
by apply nat.less_than_or_equal.rec h0; exact h1

def decreasing_induction {P : ℕ → Sort*} (h : ∀ n, P (n + 1) → P n) {m n : ℕ} (mn : m ≤ n)
(hP : P n) :
  P m :=
le_rec_on mn (λ k ih hsk, ih $ h k hsk) (λ h, h) hP
```
or
```suggestion
lemma le_induction {P : ℕ → Prop} {m}
  (h0 : P m) (h1 : ∀ n, m ≤ n → P n → P (n + 1)) :
    ∀ n, m ≤ n → P n :=
by apply nat.less_than_or_equal.rec h0; exact h1

def decreasing_induction {P : ℕ → Sort*} (h : ∀ n, P (n + 1) → P n) {m n : ℕ} (mn : m ≤ n)
  (hP : P n) :
    P m :=
le_rec_on mn (λ k ih hsk, ih $ h k hsk) (λ h, h) hP
```
Essentially, do we want some subset of the addition rules:
* Wrapped arguments are not indented
* A newline must appear after the colon if the argument list spanned multiple lines
* If a newline appears after a colon, the theorem statement gains another level of indentation

Vierkantor: Note that the goal of the pull request is descriptive: I wrote down what I understand as the current mathlib style. If we want to make changes to the conventions, let's make a poll on Zulip first.

For the record: I agree that it's almost always more readable to require a line break after `:` and `:=`, with the main exception being when the theorem is called  `bar_of_foo`, so we'd expect the conclusion line to be `foo -> bar`. If we turn `foo` into a named argument, I prefer keeping it on the same line. Something like:
```lean
lemma no_divisors_of_gcd_eq_one {x y : R}
  (h : gcd x y = 1) : ∀ z, z | x -> z | y -> is_unit z :=
sorry
```

templates/contribute/style.md
robertylewis: "type of a definition"? "declared type of a definition"?

templates/contribute/style.md
```diff
@@ -94,6 +94,8 @@ See our [documentation requirements](doc.html) for more suggestions and examples
 
 ### Structuring definitions and theorems ###
```
robertylewis, line 95:
```suggestion
### Structuring definitions and theorems
```

templates/contribute/style.md
Vierkantor:
```suggestion
These guidelines hold for declarations starting with `def`, `lemma` and `theorem`. For "theorem statement", also read "type of a definition" and for "proof" also read "definition body".
```

PR #144, opened 2020-10-07
Expand two entries in `lean.bib`

Vierkantor:
I saw that my `ring_exp` paper only had a short description in `lean.bib`, so I added information from dblp to the two entries that could be expanded.


bryangingechen:
Thanks!

PR #147, opened 2020-10-14
Some style tweaks

### LINKED ISSUE #127: Syntax highlighting
Pygments 2.7.0 was released so we should figure out how to syntax-highlight pages like https://leanprover-community.github.io/extras/calc.html

gebner:
Fixes #127.

### DIFF
--- a/templates/mathlib_stats.html
+++ b/templates/mathlib_stats.html
@@ -7,7 +7,7 @@
 
     <h2>Declaration Counts</h2>
 
-    <table class="table table-borderless mt-4" style="font-size: 1.3rem">
+    <table class="table table-borderless table-sm mt-4" style="font-size: 1.25rem">
         <tr>
             <th>Definitions</th>
             <th>Theorems</th>
--- a/templates/mathlib_stats.html
+++ b/templates/mathlib_stats.html
@@ -7,7 +7,7 @@
 
     <h2>Declaration Counts</h2>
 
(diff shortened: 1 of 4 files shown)


templates/mathlib_stats.html
```diff
@@ -6,20 +6,19 @@
...
-    </div>
-    <div class="col-md-4">
-      <h3>Meta declarations</h3>
-      <h4>{{ num_meta }}</h4>
-    </div>
-  </div>
+
+    <table class="table table-borderless mt-4" style="font-size: 1.3rem">
+        <tr>
+            <th>Definitions</th>
```
robertylewis, line 12: These headers are now styled sans-serif in my browser, in contrast to the section headers. Was that intentional? 
image

gebner: Yes, I found it a bit confusing that there were three levels of big, bold, serif headings.  The sans serif rendering looks a bit off in chromium (that's the screenshot, right?), I could tweak the font size/weight a bit if that bothers you.

robertylewis: It's Chrome, but yeah. Can't say I feel very strongly either way, but yeah, maybe a little tweaking would help. Do you think there's too much space between the headers and the numbers?

gebner: Ok, I've reduced the font size and space between the rows a bit.  I've also added a bit of indentation, which really helps a lot.

PR #149, opened 2020-10-27
show contributor count on stats page

robertylewis:


templates/mathlib_stats.html
```diff
@@ -5,18 +5,18 @@
...
 
     <table class="table table-borderless table-sm mt-4 ml-5" style="font-size: 1.25rem">
         <tr>
             <th>Definitions</th>
             <th>Theorems</th>
-            <th>Meta declarations</th>
+            <th>Contributors</th>
         </tr>
         <tr>
             <td>{{ num_defns }}</td>
```
eric-wieser, line 17: Possibly
```suggestion
            <td>{{ num_defns }} (+ {{num_meta}} meta)</td>
```
Could link `meta` to https://leanprover-community.github.io/extras/tactic_writing.html, although it's not a perfect match.

robertylewis: IMO this is more clutter than it's worth, not sure the meta number is particularly interesting to most people.

bryangingechen: We could hide more detailed statistics inside a `<details>` dropdown.

bryangingechen: Probably something for another PR though.

PR #150, opened 2020-11-11
Added suggestion to uninstall Windows Store version of Python

shingtaklam1324:
The Windows Store installation of Python seems to cause issues with installing `leanproject`. I think even with the `cp` line there the default for `python3` is still the store version of Python, and running `python3 -m pip` gives an error message saying `Permission Denied`

### DIFF
--- a/templates/install/windows.md
+++ b/templates/install/windows.md
@@ -36,6 +36,7 @@ In order to use mathlib supporting tools, you need to [get python](https://www.p
 
 ### Get Python
 
+* If you have the Windows Store version of Python, please uninstall it first.
 * Download the latest version of python [here](https://www.python.org/downloads/).
 * Run the downloaded file (`python-3.x.x.exe`)
 * Check `Add Python 3.x to PATH`.


PatrickMassot:
Can any Windows user comment on this PR? @fpvandoorn or @sgouezel maybe?

shingtaklam1324:
Apologies, I should have added some context to this PR. I made this PR when a user on the Discord said they couldn't follow the Windows installation instruction, due to the `source ~/.bash_profile` failing, and after restarting (since the `source` failed) `leanproject` still wasn't found. When I asked the user to try `pip list | grep mathlib`, a permission denied error came up. I then found out that the user was using Python from the Windows Store, and I suggested to them to uninstall and follow the Windows "Get Python" instructions.

To be honest, I think there may be a way to get it to work properly with the Windows Store version of Python, but I'm not sure how. Part of the issue is that the Windows Store version installs in a directory with restrictive access rights, which may cause issues?

https://stackoverflow.com/questions/56974927/permission-denied-trying-to-run-python-on-windows-10 for reference (it might be a Git Bash bug as well?)

shingtaklam1324:
Update - https://stackoverflow.com/questions/56974927/permission-denied-trying-to-run-python-on-windows-10 for reference

I think it might in fact be a Git Bash bug? However when I just tried, I can't reproduce the failure to install `mathlibtools` with the Windows Store version of Python :/

fpvandoorn:
The installation script already *tries* to deal with this issue under the `which python` step. It has the same suggestion that was also mentioned in the Stackoverflow thread (turn toggles off under `manage app execution aliases` ).
Do you know if the user followed this step?

Given that you could follow the installation with the Windows Store version of Python makes me suspicious of merging this PR.

shingtaklam1324:
Right, looking at it now the rest of the installation instructions does handle this. It was over a month ago so I don't really have a way to check whether the user followed that step. Given that no one else has reported this issue, I will close this PR.

PR #154, opened 2020-12-29
contribute/style: Convention for placing hypotheses before colon

MohamedAlFahim:
Based off a discussion at https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Explicit.20hypothesis.20vs.20implication.20hypothesis.20convention.
This convention could nevertheless be expanded upon.

### DIFF
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -265,6 +265,18 @@ is preferred over
 example (n : ℝ) : n > 1 → n > 0 := λ h, by linarith
 ```
 
+and
+
+```lean
+example (n : ℕ) : n ≥ 0 := dec_trivial
+```
+
+is preferred over
+
+```lean
+example : ∀ (n : ℕ), n ≥ 0 := λ n, dec_trivial
+```
+
 ### Binders ###
 
 Use a space after binders:
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -394,8 +394,9 @@ short definitions, or to group together a definition and notation.
 
 ### Hypotheses Left of Colon ###
 
-Generally, having hypotheses to the left of the colon is preferred
-over having hypotheses in implications, when possible. For instance:
+Generally, having arguments to the left of the colon is preferred
+over having arguments in universal quantifiers or implications, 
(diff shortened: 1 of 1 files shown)


templates/contribute/style.md
```diff
@@ -392,6 +392,21 @@ We generally use a blank line to separate theorems and definitions,
 but this can be omitted, for example, to group together a number of
 short definitions, or to group together a definition and notation.
 
+### Hypotheses Left of Colon ###
```
bryangingechen, line 395: I think this should go earlier in the document, maybe right before the `### Binders ###` section.

Otherwise LGTM.

templates/contribute/style.md
```diff
@@ -392,6 +392,21 @@ We generally use a blank line to separate theorems and definitions,
...
+over having hypotheses in implications, when possible. For instance:
+
+```lean
+example (n : ℝ) (h : n > 1) : n > 0 := by linarith
+```
+
+is preferred over
+
+```lean
+example (n : ℝ) : (n > 1) → (n > 0) := λ h, (by linarith)
```
fpvandoorn, line 407:
```suggestion
example (n : ℝ) : n > 1 → n > 0 := λ h, by linarith
```

templates/contribute/style.md
```diff
@@ -392,6 +392,21 @@ We generally use a blank line to separate theorems and definitions,
 but this can be omitted, for example, to group together a number of
 short definitions, or to group together a definition and notation.
 
+### Hypotheses Left of Colon ###
+
+Generally, having hypotheses to the left of the colon is preferred
+over having hypotheses in implications, when possible. For instance:
```
fpvandoorn, line 398:
```suggestion
Generally, having arguments to the left of the colon is preferred
over having arguments in universal quantifiers or implications, 
if the proof starts by introducing these variables. For instance:
```
I think it's good to make this rule only apply if the proof starts by introducing the variable, (e.g. by `λ` or by `intro`). It's fine to keep the argument on the right if the first step of the proof is something else, e.g. induction.

MohamedAlFahim:
I will add a universal quantifier example too.

PR #156, opened 2021-01-03
complete rewrite of simp.md

kbuzzard:
I rewrote the (quite crappy) simp doc. It's now a bit less crappy. I wrote the original version when I was learning how to use `simp` and had no idea what was important.

### DIFF
--- a/templates/extras/simp.md
+++ b/templates/extras/simp.md
@@ -1,111 +1,254 @@
 # Simp
 
-The `simp` tactic works by applying a conditional term rewriting system to try and prove, or at
-least simplify, your goal. What this basically means is that `simp` is equipped with a list of
-lemmas (those tagged with the `simp` attribute), many of which are of the form `X = Y` or `X iff Y`,
-and attempts to match subterms of the goal with the left hand side of a rule, and then replaces the
-subterm with the right hand side. The system is conditional in the sense that lemmas are allowed to
-have preconditions (`P -> (X = Y)`) and in these cases it will try and prove the precondition using
-its simp lemmas before applying `X = Y`.
+## Overview
 
-You can watch `simp` in action by using `set_option trace.simplify true` in your code. For example
+In this document we will explain basic usage of Lean's simplifier,
+`simp`, and the related tactic `dsimp`.
+
+## Introduction
+
+Lean has a "simplifier", called `simp`. This is a tactic whose job,
+unfortunately for beginners, is *not* to prove an arbitrary simple
+statement. Lean's simplifier is what is called a "conditional term
+rewriting system". Rewriting is the act of changing a goal of the form
+`P(A)` into `P(B)`, given a hypothesis of the form `A = B` or `A ↔
+B`. The simplifier's job is to try and prove, or at least simplify,
+goals or hypotheses, by using repeated rewrites. Here is an example
+(using `mathlib`).
 
 ```lean
-namespace hidden
+import algebra.group.defs
 
-definition cong (a b m : ℤ) : Prop := ∃ n : ℤ, b - a = m * n
+variables (G : Type) [group G] (a b c : G)
 
-notation a ` ≡ ` b ` mod ` m  := cong a b m
-set_option trace.simplify true
-theorem cong_refl (m : ℤ) : ∀ a : ℤ, a ≡ a mod m :=
+example : a * a⁻¹ * 1 * b = b * c * c⁻¹ :=
 begin
-intro a,
-unfold cong,
-existsi (0:ℤ),
-simp
+  simp
 end
-
-end hidden
 ```
 
-If you do this exercise you will discover firstly that `simp` spends a lot of its time trying random
-lemmas and then giving up very shortly afterwards, and also that the `unfold` command is also
-underlined in green -- Lean seems to apply `simp` when you do an `unfold` as well (apparently
-`unfold` just asks `simp` to do its dirty work for it -- `unfold X` is close to `simp only [X]`).
-
-If you only want to see what worked rather than all the things that didn't, you could try
-`set_option trace.simplify.rewrite true`.
+How would a human solve that goal? They would notice that `a * a⁻¹ =
+1`, that `1 * 1 = 1` and so on, until they had simplified the example
+to `b = b`, which is obviously true. This is also what the simplifier
+is doing. Indeed, if you add `set_option trace.simplify.rewrite true`
+above the example, then a squiggly blue underline will appear under
+`simp` (in VS Code) and clicking on this will show you the sequence of
+rewrites that `simp` discovered.
 
 ## Simp lemmas
 
-In case you want to train `simp` to use certain extra lemmas (for example because they're coming up
-again and again in your work) you can add new lemmas for yourself. For example in mathlib in
-`algebra/ring.lean` we find the line
+So how did Lean's simplifier know that `a * a⁻¹ = 1`? It is because
+the corresponding lemma in `algebra.group.defs` is tagged with the
+`simp` attribute:
 
 ```lean
-@[simp] theorem ne_zero (u : units α) : (u : α) ≠ 0
+@[simp] lemma mul_right_inv (a : G) : a * a⁻¹ = 1 := ...
 ```
 
-This lemma is then added to `simp`'s armoury. Note several things however.
-
-1) It might not be wise to make a random theorem into a simp lemma. Ideally the result has to be of
-a certain kind, the most important kinds being those of the form `A=B` and `A↔B`. Note however that
-if you want to add `fact` to `simp`'s weaponry, you can prove
+This `@[simp]` at the beginning just "tags" `mul_right_inv` with the
+`simp` attribute. We call lemmas tagged in this way "`simp`
+lemmas". Here are some examples of `simp` lemmas in mathlib:
 
 ```lean
-@[simp] lemma my_lemma : fact ↔ true
+@[simp] theorem nat.dvd_one {n : ℕ} : n ∣ 1 ↔ n = 1 := ...
+@[simp] theorem mul_eq_zero {a b : ℕ} : a * b = 0 ↔ a = 0 ∨ b = 0 := ...
+@[simp] theorem list.mem_singleton {a b : α} : a ∈ [b] ↔ a = b := ...
+@[simp] theorem set.set_of_false : {a : α | false} = ∅ := ...
 ```
 
-(and in fact more recent versions of Lean do this automatically when you try to add random theorems
-to the simp dataset).
+When the simplifier is attempting to simplify a term `T` it looks
+through the `simp` lemmas known to the system at that time, in a
+sensible way (there are over ten thousand of them currently in
+`mathlib`!), and if it runs into a lemma of the form `A = B` or `A ↔
+B`, for which `A` shows up in `T`, it rewrites the lemma at `T` to
+replace it with a `B`, and starts again.
 
-2) If you are not careful you can add a bad simp lemma of the form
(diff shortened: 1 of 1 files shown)


templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
-lemmas and then giving up very shortly afterwards, and also that the `unfold` command is also
-underlined in green -- Lean seems to apply `simp` when you do an `unfold` as well (apparently
-`unfold` just asks `simp` to do its dirty work for it -- `unfold X` is close to `simp only [X]`).
-
-If you only want to see what worked rather than all the things that didn't, you could try
-`set_option trace.simplify.rewrite true`.
+How would a human solve that goal? They would notice that `a * a⁻¹ =
+1`, that `1 * 1 = 1` and so on, until they had simplified the example
+to `b = b`, which is obviously true. This is also what the simplifier
+is doing. Indeed, if you add `set_option trace.simplify.rewrite true`
```
bryangingechen, line 33: Could you mention `set_option trace.simplify true` as well? I think I've referred a few people to this page on Zulip for these trace options.

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
+@[simp] theorem set.set_of_false : {a : α | false} = ∅ := ...
 ```
 
-This lemma is then added to `simp`'s armoury. Note several things however.
+When the simplifier is attempting to simplify a term `T` it looks
+through the `simp` lemmas known to the system at that time, in a
+sensible way (there are over ten thousand of them currently in
+`mathlib`!), and if it runs into a lemma of the form `A = B` or `A ↔
+B`, for which `A` shows up in `T`, it rewrites the lemma at `T` to
+replace it with a `B`, and starts again.
```
bryangingechen, line 64: Worth mentioning here (with an example, maybe) that `simp` works "inside-out".

bryangingechen (commented):
Thanks! This is definitely an improvement.

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
+          Lean at that time.
+
+`simp [h1, h2]` -- uses all `simp` lemmas and also `h1` and `h2`
+                   (which can either be local hypotheses or other lemmas which
+		   are not tagged `simp` for some reason).
+
+`simp [← h]` -- uses all `simp` lemmas, and also `h : A = B` but in
+                 the form `B = A` (so `simp` rewrites `B`s to `A`s)
+
+`simp [-h]` -- stops `simp` from using lemma `h`.
```
digama0, line 112: Maybe this should say `simp [-thm]`, to make it clear that what goes there is not a local hypothesis but a defined theorem.

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
-[`squeeze_simp`](https://leanprover-community.github.io/mathlib_docs/tactics.html#squeeze_simp%20/%20squeeze_simpa%20/%20squeeze_scope).
+1) Use `squeeze_simp` instead of `simp`. This produces output of the
+form `simp only [h1, h2, ..., hn]`, which can be used instead of
+`simp` to perform the same task. `simp only` is much safer to use in
+the middle of a proof, because the simplifier is now constrained to
+only using the lemmas in the list, so will not be affected by changes
+to the `simp` set (the collection of lemmas tagged `simp`).
 
-## How to use simp better
+2) `simpa`. If `simp` turns your goal into `h`, then you can write
```
digama0, line 214:
```suggestion
2) `simpa`. If `simp` turns your goal into `P`, then you can write
```

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
+```
 
-1) `simp only [H1, H2, H3]` uses only lemmas `H1`, `H2`, and `H3` rather than `simp`s full
-collection of lemmas. Whyever might one want to do this in practice? Because sometimes `simp`
-simplifies things too much -- it might unfold things that you wanted to keep folded, for example.
-Another reason is that using `simp only` can speed up slow `simp` calls significantly.
+which can be a very useful way of tidying up a goal. It can be safely
+used in the middle of a proof because of the `only`. This command
+tidies up lambdas (it will turn `(λ x, f x) 37` into `f 37`, and
+`{to_fun := f, ...}.to_fun` into `f`.
```
digama0, line 240:
```suggestion
`{to_fun := f, ...}.to_fun` into `f`).
```

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
 
-3) `simp * at *`. This simplifies everything in sight. Use if life is getting complicated.
+`simp {single_pass := tt}` -- this `single_pass` is a config option,
+one of around 16 at the time of writing. One can use `single_pass` to
+avoid loops which would otherwise occur.
 
-4) `simp {single_pass := tt}` -- this `single_pass` is a config option, one of around 16 at the time of writing. One can use `single_pass` to avoid loops which would otherwise occur; for example `nat.gcd_def` is an equality with `gcd` on both the left and right hand side, so `simp [nat.gcd_def]` is risky behaviour whereas `simp [nat.gcd_def] {single_pass := tt}` is not. As you can imagine, `simp only [h] {single_pass := tt}` here makes simp behave pretty much like `rw h`.
+Searching for `structure simp_config` in the file
+`init/meta/simp_tactic.lean` in core Lean reveals other config
+options, not all of them documented.
```
digama0, line 250: Would it be helpful to document all of them in this file? I can probably give short descriptions of all of them with some spelunking. Most of them aren't very relevant for the average user.

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
+`simp [-h]` -- stops `simp` from using lemma `h`.
+
+`simp *` -- uses all `simp` lemmas and also all current local
+            hypotheses to try to simplify the goal.
+
+`simp at h` -- tries to simplify `h` using all `simp` lemmas.
+
+`simp [h1] at h2 ⊢` -- tries to simplify both `h2` and the goal using
+                       `h1` and all `simp` lemmas (note: type `⊢` with
+                       `\|-` in VS Code).
```
jcommelin, line 121:
```suggestion
                       `\|-` or `\vdash` in VS Code).
```

templates/extras/simp.md
bryangingechen: That'd be great! There is a list of them at the bottom of the file, but very few of them have descriptions.

templates/extras/simp.md
kbuzzard: Done.

templates/extras/simp.md
kbuzzard: I don't know what this means. Can you sketch the idea? I still have no idea how simp works, I just now know what it does.

templates/extras/simp.md
kbuzzard: Thanks

templates/extras/simp.md
kbuzzard: Done

templates/extras/simp.md
kbuzzard: Thanks.

kbuzzard:
Sorry I left this for a few days -- normal work / home jobs + LT2021 was a bit much! Currently I need advice on the "outside in" comment above, and also I am a bit concerned that my comments on simp normal form.

More precisely, I use (n : nat) > 0 as the first example with simp normal form and I've since discovered that the situation is more complicated than I had envisaged (https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.60simp.60.20normal.20form/near/221451173). For example Floris' comment about how `_ \ne 0` should be "standard" but it wasn't clear that it should be the "simp normal form" just indicate that again my understanding of what is going on is too poor to be able to write the docs properly (I don't know the difference). If nobody can explain to me what the difference is then I might just switch to another example. I was hoping for an example which would be comprehensible to both mathematicians and computer scientists.

An example which I think I do understand is membership of subgroups: 

```
@[simp, norm_cast, to_additive]
lemma mem_coe {K : subgroup G} {g : G} : g ∈ (K : set G) ↔ g ∈ K := iff.rfl
```

tells us that the simp normal form is `g ∈ K`, and simp lemmas such as

```
@[simp, to_additive] theorem inv_mem_iff {x : G} : x⁻¹ ∈ H ↔ x ∈ H :=
⟨λ h, inv_inv x ▸ H.inv_mem h, H.inv_mem⟩
```

are fine and then you explain why you now don't also need analogous lemmas mentioning stuff like  `x⁻¹ ∈ \u H` . Bit I was concerned these were a bit mathematical for computer scientists. I think this is quite a good example though.

templates/extras/simp.md
kbuzzard: Thanks

templates/extras/simp.md
bryangingechen: [Here's](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Simp.20normal.20form/near/188829903) a post by Gabriel that I starred in Zulip about this:

> The simplifier simplifies terms inside-out, that is, it first simplifies the arguments of a function before simplifying the function. Hence a simp lemma is only ever used to rewrite terms where the arguments are already fully simplified (= in simp-normal form). Corollary: if the left-hand side of your simp lemma has an argument that is not fully simplified, then this simp lemma is never used to rewrite a term by the simplifier and is hence useless.

jcommelin:
@digama0 could you take a look at this one?

kbuzzard:
I've added the "inside-out" comments, finally. I am not completely happy with my "simp normal form" explanation; should I just remove all the 0 < n stuff?

templates/extras/simp.md
```diff
@@ -1,111 +1,265 @@
...
+```
+@[simp] lemma mem_coe {a : α} {s : finset α} : a ∈ (s : set α) ↔ a ∈ s := ...
+```
+
+will attempt to turn any occurrences of `a ∈ (s : set α)` into the
+correct normal form.
+
+Because the simplifier works from the inside out, simplifying arguments
+of a function before simplifying the function, a `simp` lemma should
+always have all of its arguments already in simp-normal form. Mathlib's
```
robertylewis, line 192: The issue is with the (arguments of the) LHS of the simp lemma, not the arguments to the lemma itself. For instance, if you have @[simp] lemma fg (cond : P) : f (g 0) = 0 but g 0 already simplifies to something else, fg will never fire. I read "arguments of fg" as referring to hypotheses like cond, but it doesn't matter if cond is in simp-normal form.
 
```suggestion
the arguments of the left hand side of a simp lemma 
should already be in simp-normal form. Mathlib's
```
?
You could add an example if you want.

kbuzzard: Thanks. It's clear that I'm not really the best person to be writing this page, however I am very much motivated to get something better than what we currently have because students can ask direct questions such as "how can I learn about `simp`?". Thanks all of you for the help you've given.

templates/extras/simp.md
```diff
@@ -1,111 +1,254 @@
...
+file, even though they are true by definition, because they are
+important `simp` lemmas. The very slick proof in that file that the
+complex numbers are a commutative ring uses `simp` at a key moment to
+reduce goals such as `(z + w).re = (w + z).re` into `z.re + w.re =
+w.re + z.re`, which is a statement about real numbers which the `ring`
+tactic can solve.
+
+## Basic usage 
+
+`simp` -- tries to simplify the goal using all `simp` lemmas known to
```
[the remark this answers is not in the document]
kbuzzard: I have removed the indentation. I don't really know anything about markdown, web pages or anything. Sorry. I've also removed the dashes -- I decided they weren't really adding anything.

robertylewis:
Thanks very much for this, Kevin, it's a huge improvement! I'm gonna merge this now and we can keep tweaking if needed.

PR #162, opened 2021-01-26
add punctuation to paper list; fix name

robertylewis:


lean.bib
```diff
@@ -486,7 +486,7 @@ @Article{	  SelsamDeMoura16
 }
 
 @Article{	  SelsamHudonDeMoura20,
-  author	= {Selsam, Daniel and Hudon, Simon and de Moura, Leonardo},
+  author	= {Selsam Daniel and Hudon, Simon and {de Moura}, Leonardo},
```
gebner, line 491: My bibtex is a bit rusty, but isn't the comma here necessary?

robertylewis: Good catch, thanks. That wasn't intentional.
```suggestion
  author	= {Selsam, Daniel and Hudon, Simon and {de Moura}, Leonardo},
```

PR #165, opened 2021-02-14
refactor: port mistletoe + MathJax changes from doc-gen

bryangingechen:
This ports the changes to the mistletoe CustomHTMLRenderer class from leanprover-community/doc-gen#110 that are needed to make it work well with MathJax:

- moved all `CustomHTMLRenderer` code to `mistletoe_renderer.py`
- added helper module `mathjax_editing.py` (only called by `mistletoe_renderer`)
- added test file `latex.md` (will be viewable at https://leanprover-community.github.io/latex.html after this is deployed, but should not be linked to from anywhere else)

One non-MathJax-related change in behavior in `CustomHTMLRenderer` is that the syntax highlighting now defaults to `lean`, rather than attempting to guess a language. This makes it agree with both Zulip and `doc-gen`. I thus added info strings to the markdown files that needed them.


data/community.md
```diff
@@ -8,38 +8,38 @@ data scientists and software engineers working in industry.
...
 #check my_nat
 ```
 ````
 
-You can use LaTeX using `$$` to enclose inline LaTeX and 
-````
-```latex
+You can use LaTeX using `$$` to enclose inline LaTeX and
+````md
+```math
```
bryangingechen, line 42: This, at least, was wrong previously.

gebner:
Is this consistent with `doc-gen`?

bryangingechen:
> Is this consistent with `doc-gen`?

It will be now that https://github.com/leanprover-community/doc-gen/pull/110 is merged. (I will try to backport some of the MathJax related changes here next.)

bryangingechen:
Actually, let me rework this more along the lines of leanprover-community/doc-gen#110. Namely, I want to see if I can factor out the markdown processing into another module like I did there.

gebner:
`__pycache__` should probably be in gitignore.

PR #170, opened 2021-02-20
fix(theories/topology.md) update set complement and set subtraction notation

dwrensha:
Fixes an example that @jcreedcmu noticed to be broken.

Set complement notation changed in https://github.com/leanprover-community/lean/pull/367 and https://github.com/leanprover-community/mathlib/pull/3212.

I suspect something similar happened with set subtraction, but I have been unable to find
the exact point in history.


bryangingechen:
Thanks!

PR #171, opened 2021-02-22
port relevant changes from doc-gen

bryangingechen:


robertylewis:
@bryangingechen I just noticed that the website build had gotten a lot slower and traced it back here. Markdown rendering goes from <1 sec to the better part of a minute with this commit. Was that expected? I guess it's expensive to check for urls everywhere?

It's not important, the speed of the website build doesn't matter. I was just surprised.

bryangingechen:
Oh, whoops, I guess I should have noticed. If the build gets too slow we can take this feature out. It looks like it maybe slowed down `doc-gen` by about 8 minutes as well.

PR #173, opened 2021-03-01
try to fix markdown rendering errors

eric-wieser:
This doesn't render correctly on the web version, even if it works fine on github


bryangingechen:
I just tested this out locally and unfortunately the backslashes appear in the rendered output. I think there's most likely an issue with the MathJax parsing code. I'll try to take a look at this in a few days.

bryangingechen:
This is fixed now: https://leanprover-community.github.io/extras/tactic_writing.html#marios-monadic-symbols-cheat-sheet

The issue was that the regex wasn't detecting math in backticks properly. I think I somehow copy and pasted the regex from an earlier version of the code in https://dev.sstatic.net/Js/mathjax-editing.en.js . Anyways, it seems to work after I updated it to what StackExchange is using now.

Thanks for reporting!

PR #174, opened 2021-03-01
Show the pattern matching `else` syntax

eric-wieser:


eric-wieser:
@digama0 or @bryangingechen, mind taking a look?

bryangingechen (approved):
I didn't know about this syntax either!

digama0:
I would avoid comparing `<|>` to `|` as they are really quite different syntactically and semantically. `<|>` is a regular binary infix operator that fails if both of the inputs fail, while `|` is a part of the `a <- m | m2,` syntax inside a do block, and the result of `m2` goes to the return of the tactic rather than bound into `a`.

eric-wieser:
Sure, they're quite different operators - but I suspect lots of people reach for `<|>` when what they really want is `|`, so I think it makes sense to keep them together. I'll see if I can reword slightly.

templates/extras/tactic_writing.md
```diff
@@ -171,9 +171,30 @@ the left and right hand sides of the current goal. It also uses the
...
+where execution only proceeds to the next line of the `do` block if the pattern matches, and otherwise
+executes the expression after the `|`:
+```lean
+meta def trace_goal_is_eq : tactic unit :=
+do `(%%l = %%r) ← tactic.target | tactic.trace "Goal is not an equality",
+   tactic.trace $ "Goal is equality between " ++ (to_string l) ++ " and " ++ (to_string r)
+```
+
+If the `|` is ommitted, then the tactic fails if the pattern does not match.
+We can catch this failure using the `orelse` combinator mentioned ealier, but not that by
```
eric-wieser, line 190:
```suggestion
We can catch this failure using the `orelse` combinator mentioned ealier, but note that by
```

eric-wieser:
@digama0, does this look better?

digama0 (approved):

PR #177, opened 2021-03-15
style(meet): Use bootstrap media rather than cards

eric-wieser:
This makes the page layout a little denser.

Implemented as described at https://getbootstrap.com/docs/4.6/components/media-object

----

Before / after, showing the change applied to just the top two:

image

### DIFF
--- a/templates/meet.html
+++ b/templates/meet.html
@@ -4,7 +4,7 @@
 
         {{ community | md }}
 
-  <div class="row mt-3">
+  <div class="row mt-5">
         {% for maintainer in maintainers %}
 		<div class="col-md-6">
 			<div class="media mb-3">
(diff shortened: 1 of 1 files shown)


gebner:
I like that the names are aligned with the pictures now.  Can you increase the margin to the paragraph a bit?

gebner:
The vertical distance to the paragraph above.  The distance between Jeremy and Anne is ok.

templates/meet.html
```diff
@@ -7,20 +7,12 @@
   <div class="row mt-3">
```
eric-wieser, line 7:
```suggestion
  <div class="row mt-5">
```

eric-wieser:
Looks like @robertylewis's new non-square picture makes the layout slightly inconsistent, but it's not a big deal.

robertylewis:
Oops, I'll fix that soon. Figured it was okay since Simon also has a non-square picture, but he has a transparent border.

PR #180, opened 2021-04-10
leanproject.md: link "olean files"

waldyrious:
Add contextual information to the first occurrence of the term.

Also add missing blank lines around code blocks.

### DIFF
--- a/templates/leanproject.md
+++ b/templates/leanproject.md
@@ -22,6 +22,7 @@ project is to run:
 ```text
 leanproject get tutorials
 ```
+
 You can specify a git branch name `my_branch` by appending
 `:my_branch` at the end of the specified name (without space).
 By default this branch should be an existing branch.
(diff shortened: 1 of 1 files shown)


templates/leanproject.md
```diff
@@ -34,18 +35,22 @@ command.
...
 the current folder. This new project will be using the latest version of
 Lean compatible with mathlib, and include a pre-built mathlib.
 
 ### Building a project
 
-Only mathlib itself comes with pre-built olean files. In order to build
-oleans in a project (which is needed for every non-trivial project in
-order to get decent interactive Lean speed), you can use:
+Only mathlib itself comes with pre-built
+[olean files](https://github.com/leanprover/tutorial/blob/master/05_Interacting_with_Lean.org#projects).
```
bryangingechen, line 50: I'm uncomfortable with linking to this old tutorial which is out of date in many ways. I think it might be better just to say here that oleans are binary files that Lean generates when checking a `.lean` file.

Maybe others feel differently though.

waldyrious: Is the information from that section alone something that could be repurposed to this page? Or is it out out date as well? I find it unfortunate that the information on the projects' internals isn't easier to find. This page ("Using leanproject") seems to be a good place for that kind of information, so if it is still valid, I'd be happy to copy it over here so it won't remain tucked away in the contents of an old tutorial.

bryangingechen: That's a good question. Do you mind joining [our Zulip chat](https://leanprover.zulipchat.com/) and making a thread about this? I think folks there may have a better sense of what would be good to include from that page. (In terms of stuff that's outdated, certainly all the `.hlean` stuff is no longer relevant, and I think `lean -o` no longer works either. Most of the other stuff looks ok to me though.)

waldyrious: Sure! It may take me a few days to find the bandwidth to engage in that conversation (maybe next weekend?), but I'll get to it.

semorrison:
I'm going to close this now, we're going to be completely replacing this page shortly in any case.

waldyrious:
Thanks for the update, and apologies for not having had the time to pursue this further.

PR #184, opened 2021-04-23
lean.bib: add paper on Dedekind domains and class groups

Vierkantor:
Add our paper "A formalization of Dedekind domains and class groups of global fields", to be presented at ITP 2021, to the bibliography file. Once the paper is published, I'll update the reference to a full `InProceedings`.


bryangingechen:
Thanks!

PR #185, opened 2021-04-29
feat(theories): update linear algebra theory docs

Vierkantor:
Several refactors have made the linear algebra theory docs outdated. This PR updates them to the current state of mathlib.

The `bundled_basis` refactor that I am working on should not re-outdate these docs, by the way.


jcommelin (approved):
Thanks!

PR #188, opened 2021-06-14
Move current installation instructions to a lean3 subdirectory

agjftucker:
My idea is to move the current pages about getting started / installation to a new lean3 subdirectory. That way, a project written today with `leanproject` instructions and a link to `lean3/get_started` in its README should continue to work without modification even after the default package manager has become something else.

Sorry my web skills are rubbish! When I brought this up on Zulip it was suggested that a "lightweight redirect" might be appropriate. I'm not sure whether this PR answers that description.

I think for this to work, the master branch might need the `lean3` and `lean3/install` subdirectories to be created by hand. That's what I had to do locally in my `build` folder.


jcommelin:
Thanks for the PR. I agree that it's a good idea to move these pages well before lean 4 becomes the default.

PatrickMassot:
I'm not so sure. Anything that may lead people to think Lean4 could already be the default sound very confusing to me. We already see new zulip users confused about that pretty often since January.

semorrison:
This has finally become redundant, as I've made these changes now...

PR #194, opened 2021-08-11
Update the macOS detailed install instructions to recommend Homebrew.

Julian:
Corresponds with leanprover-community/mathlib-tools#108.

### DIFF
--- a/templates/install/macos.md
+++ b/templates/install/macos.md
@@ -9,6 +9,8 @@ There is a [video walkthrough](https://www.youtube.com/watch?v=NOGWsCNm_FY) of t
 
 ## Installing Lean and mathlib
 
+### Intel Macs
+
 Here we will discuss the fast way, assuming a lot of trust from you. It
 will install Lean, with supporting tools `elan` and `leanpkg`,
 the supporting tool `leanproject` for Lean's mathematical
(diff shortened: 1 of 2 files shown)


FrederickGeek8:
Does it make sense to add the steps
```bash
elan toolchain install stable
elan default stable
```
since they are _also_ executed by the install script and necessary for a first-time install?

Julian:
Yes! Good point, will add.

Julian:
@FrederickGeek8 care to leave an lgtm here too if you agree with these changes?

@semorrison or @bryangingechen perhaps one of you two are OK with this at this point as well?

bryangingechen:
I'd like to test this, but I installed `elan` and `mathlibtools` manually a while back. Is it worth adding instructions on how to cleanly uninstall those and switch to the `homebrew` method?

bryangingechen:
You mentioned on Zulip that these instructions work for M1 Macs as well? Is that worth mentioning explicitly somewhere in the docs?

Julian:
Thanks!

> I'd like to test this, but I installed elan and mathlibtools manually a while back. Is it worth adding instructions on how to cleanly uninstall those and switch to the homebrew method?

Definitely, will do in a bit (let me know if you see a decent place to add this, otherwise the best I see at the minute is tacking it to the bottom of one of these two pages).

> You mentioned on Zulip that these instructions work for M1 Macs as well?

Only the `leanproject` / mathlibtools part unfortunately. Provisioning lean / elan still will not, so it's only half this page that's relevant. I filed leanprover-community/mathlib-tools#107 with an idea for how to automate M1s as well, certainly would love your thoughts there -- either way I suspect an M1-specific section may be necessary for the detail page here. I'd personally probably not block updating the page on that given I have so little time to commit at the minute to getting that ticket done (even if you agree with the proposal), but obviously happy to do so if you think it's best.

bryangingechen:
> Only the `leanproject` / mathlibtools part unfortunately. Provisioning lean / elan still will not, so it's only half this page that's relevant. I filed [leanprover-community/mathlib-tools#107](https://github.com/leanprover-community/mathlib-tools/issues/107) with an idea for how to automate M1s as well, certainly would love your thoughts there -- either way I suspect an M1-specific section may be necessary for the detail page here. I'd personally probably not block updating the page on that given I have so little time to commit at the minute to getting that ticket done (even if you agree with the proposal), but obviously happy to do so if you think it's best.

OK, that sounds reasonable to me. Do you think it'd be a good idea to point M1 Mac users to [the Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/M1.20macs) in the meantime?

FrederickGeek8:
@Julian lgtm :)

bryangingechen:
@Julian I tested the instructions and everything seems to work great. I'm happy to merge now if you're done, though I still think some kind of note for M1 Mac users might be helpful.

Julian:
Addressing your M1 note suggestion was on my todo before calling it done, yeah, happy to of course, though glad to hear it works and looks good!

I'll add that in a few minutes.

Julian:
OK, done!

bryangingechen:
Thanks so much!

PR #195, opened 2021-08-12
doc(templates/contribute): update doc_blame linter documentation

mans0954:
The current [doc-strings contribute guidelines](https://leanprover-community.github.io/contribute/doc.html#doc-strings) refer to the `doc_blame` and `doc_blame!` commands which no longer exist. This is confusing for newcomers.

This PR updates the documentation to refer to the linters which replace these commands, and explains how they may be used by the reader who may not be familiar with the `lint` command.


PatrickMassot:
Good catch, thanks!

PR #196, opened 2021-08-20
docs(templates/extras/simp.md): edited, added table of simp options

kmill:
Added a table of all the simp/dsimp options, with descriptions of most of them.  Edited things in general for clarity.


templates/extras/simp.md
```diff
@@ -2,19 +2,31 @@
...
+Lean has a "simplifier", called `simp`, that consults a database of
+facts called *`simp` lemmas* to (hopefully) simplify hypotheses and
+goals. The simplifier is what is known as a *conditional term
+rewriting system*: all it does is repeatedly replace (or *rewrite*)
+subterms of the form `A` by `B`, for all applicable facts of the form
+`A = B` or `A ↔ B`.
+
+The simplifier mindlessly rewrites until it can rewrite no more.  The
+`simp` lemmas are all oriented, with left-hand sides always being
+rewritten by right-hand sides, and never vice versa. The database of
```
semorrison, line 20:
```suggestion
replaced by right-hand sides, and never vice versa. The database of
```

templates/extras/simp.md
```diff
@@ -2,19 +2,31 @@
...
+facts is curated such that things are put into some approximation of a normal
+form, and (at least ideally) the facts are *confluent*, meaning the
+order in which the simplifier considers rewrites does not matter.
```
semorrison, line 23:
```suggestion
rewritten by right-hand sides, and never vice versa. 

Ideally, the database of facts would result in expressions being rewritten
into a normal form. 
In practice, this is often unachievable (normal forms may not exist, or there
may not exist a collection of rewrite rules that produce them),
but nevertheless we aim to approximate this ideal where possible.

Even better, we would like the database of facts to be *confluent*,
meaning the order in which the simplifier considers rewrites does not matter.
Again, while this is not always achievable, we aim to be close to confluent
where possible.
```

templates/extras/simp.md
```diff
@@ -27,15 +39,15 @@ begin
 end
 ```
 
-How would a human solve that goal? They would notice that `a * a⁻¹ =
-1`, that `1 * 1 = 1` and so on, until they had simplified the example
+How would a human solve that goal? They would notice that `a * a⁻¹ = 1`,
+that `1 * 1 = 1`, and so on, until they had simplified the example
 to `b = b`, which is obviously true.
 
 This is also what the simplifier is doing. Indeed, if you add
```
semorrison, line 46: Could I suggest you edit this paragraph so it introduces `simp?` first, and then mentions that `set_option` possibilities if you need even more detail? I think training people to use `simp?` is good. :-)

templates/extras/simp.md
```diff
@@ -63,221 +74,251 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
-The simplifier works in one direction only: if `A = B` is a
-`simp` lemma, then `simp` replaces `A`s with `B`s, but it doesn't
-replace `B`s with `A`s. Hence a `simp` lemma should have the property
-that its right hand side should be simpler than its left hand side. In
+through the `simp` lemmas known to the system at that time, and if it
+runs into an applicable lemma of the form `A = B` or `A ↔ B` for which
+`A` shows up in `T`, it rewrites the instance of `A` in `T` with `B`
+and then starts again from the beginning. Note that `simp` starts on
+innermost terms, working outward: it first simplifies the arguments of
+a function before simplifying the function. Also, `simp` contains a some
```
semorrison, line 82:
```suggestion
a function before simplifying the function. Also, `simp` contains some
```

templates/extras/simp.md
```diff
@@ -63,221 +74,251 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
-from the inside out: it first simplifies the arguments of a function before
-simplifying the function.
-
-The simplifier works in one direction only: if `A = B` is a
-`simp` lemma, then `simp` replaces `A`s with `B`s, but it doesn't
-replace `B`s with `A`s. Hence a `simp` lemma should have the property
-that its right hand side should be simpler than its left hand side. In
+through the `simp` lemmas known to the system at that time, and if it
+runs into an applicable lemma of the form `A = B` or `A ↔ B` for which
+`A` shows up in `T`, it rewrites the instance of `A` in `T` with `B`
```
semorrison, line 79:
```suggestion
`A` appears as a subexpression in `T`, it rewrites the instance of `A` in `T` with `B`
```

templates/extras/simp.md
```diff
@@ -63,221 +74,251 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
+which has almost 100 `simp` lemmas. Even though they are true by definition, theorems such as
+```lean
+@[simp] lemma
+add_re (z w : ℂ) : (z + w).re = z.re + w.re := rfl
+```
+are introduced because they give `simp` the ability to reduce expressions and then make use of pre-existing facts.
+This one, for example, converts complex addition into real addition.
+If you give `simp` permission to use commutativity of real addition, then it is able to
+automatically prove `(z + w).re = (w + z).re` through `z.re + w.re =
+w.re + z.re`, which is half of the proof that complex addition is commutative.
```
semorrison, line 120: Maybe add here something like:

```suggestion
Anytime you create a new definition in Lean, you should also write some appropriate `simp` lemmas.
This will make it easier for others to use your definition without having to unfold it.

If you are defining something which is a structure, putting the `@[simps]` attribute on the definition
will automatically create `simp` lemmas for the projections.
```

templates/extras/simp.md
kmill: I added the output of `set_option trace.simplify.rewrite true` along with what `simp?` suggests to illustrate the meaning of `simp?`.

templates/extras/simp.md
kmill: Thanks for the suggestion. I tried explaining how it's useful having simp lemmas that do partial evaluation, and how `@[simps]` can help.

templates/extras/simp.md
```diff
@@ -62,222 +100,289 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
+| `fail_if_unchanged` | `tt` | `tt` | Fail if no simplifications applied |
+| `eta` | `tt` | `tt` | Allow eta-equivalence: `(λ x, F x)` ↝ `F` |
+| `zeta`| `tt` | `tt` | Do zeta-reductions: `let x : a := b in c` ↝ `c[x/b]` |
+| `beta` | `tt` | `tt` | Do beta-reductions: `(λ x, E) y` ↝ `E[x/y]` |
+| `proj` | `tt` | `tt` | Reduce projections: `⟨a, b⟩.1` ↝ `a` |
+| `iota` | `tt` | `tt` | Reduce recursors: `nat.rec_on (succ n) Z R` ↝ `R n (nat.rec_on n Z R)` |
+| `iota_eqn` | `ff` | | Reduce using all equation lemmas generated by the equation compiler |
+| `unfold_reducible` | | `ff` | Unfold definitions with `reducible` transparency (delta-reduce) |
+| `memoize` | `tt` | `tt` | Perform caching of simps of subterms |
+| `contextual` | `ff` | | |
```
ericrbg, line 374: Mario talks about what contextual does [here](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/custom.20simp-sets/near/235615807)

kmill: Ok, added some information

templates/extras/simp.md
```diff
@@ -62,222 +100,301 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
+| `contextual` | `ff` | | Use additional `simp` lemmas based on the context of the current subexpression (see example below)  |
+| `single_pass` | `ff` | `ff` | Visit each subterm no more than once |
+| `md` | | `reducible` | Reduction mode: how aggressively constants are replaced with their definitions (`all`, `semireducible`, `instances`, `reducible`, or `none`) |
+| `max_steps` | max | max | The maximum number of steps allowed before failing |
+| `canonize_instances` | `tt` | `tt` | Replace each instance with a canonical defeq one |
+| `canonize_proofs` | `ff` |  |  |
+| `fail_if_unchanged` | `tt` | `tt` | Fail if no simplifications applied |
+| `eta` | `tt` | `tt` | Allow eta-equivalence: `(λ x, F x)` ↝ `F` |
+| `zeta`| `tt` | `tt` | Do zeta-reductions: `let x : a := b in c` ↝ `c[x/b]` |
+| `beta` | `tt` | `tt` | Do beta-reductions: `(λ x, E) y` ↝ `E[x/y]` |
```
fpvandoorn, line 369:
```suggestion
| `zeta`| `tt` | `tt` | Do zeta-reductions: `let x : a := b in c` ↝ `c[x := b]` |
| `beta` | `tt` | `tt` | Do beta-reductions: `(λ x, E) y` ↝ `E[x := y]` |
```
Substitution is usually/often written the other way around: `c[b/x]` (see e.g. the link below). The above suggestion is unambiguous notation for substitution (but maybe confusing since we also use it in `let`?
https://math.stackexchange.com/questions/273294/substitution-in-lambda-calculus

(also fix the occurrence below the table)

fpvandoorn (commented):
Could you add an example / explanation of `simpa [h] using t`?

templates/extras/simp.md
kmill: Thanks, I'm happy to replace that with less ambiguous notation.

For what it's worth, it's lifted from https://github.com/leanprover-community/lean/blob/master/library/init/meta/simp_tactic.lean#L132

kmill:
@fpvandoorn I added a section on `simpa` and gave a contrived but plausible example.

I also tried digging up what the remaining configuration options do.

templates/extras/simp.md
```diff
@@ -62,222 +100,301 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
 
-which can be a very useful way of tidying up a goal. It can be safely
-used in the middle of a proof because of the `only`. This command
-tidies up lambdas (it will turn `(λ x, f x) 37` into `f 37`), and
-structure projections (it will turn `{to_fun := f, ...}.to_fun`
-into `f`).
+> `simp` (`!`)? (`?`)? (`only`)? (`*` | `[` list of arguments `]`)? (`with` simp sets)? (`at` locations)? (`{` configuration options `}`)?
 
-## More advanced features.
+If `!` is present, it adds `iota := tt` to the configuration options.
```
digama0, line 363:
```suggestion
If `!` is present, it adds `iota_eqn := tt` to the configuration options.
```

templates/extras/simp.md
```diff
@@ -62,222 +100,346 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
+| `beta` | `tt` | `tt` | Do beta-reductions: `(λ x, a) y` ↝ `a[x := y]` |
+| `eta` | `tt` | `tt` | Allow eta-equivalence: `(λ x, f x)` ↝ `f` |
+| `zeta`| `tt` | `tt` | Do zeta-reductions: `let x := a in b` ↝ `b[x := a]` |
+| `proj` | `tt` | `tt` | Reduce projections: `prod.fst (a, b)` ↝ `a` |
+| `iota` | `tt` | `tt` | Reduce recursors: `nat.rec_on (succ n) Z R` ↝ `R n (nat.rec_on n Z R)` |
+| `iota_eqn` | `ff` | | Reduce using all equation lemmas generated by the equation compiler |
+| `unfold_reducible` | | `ff` | Unfold definitions with `reducible` transparency (delta-reduce) |
+| `memoize` | `tt` | `tt` | Perform caching of simps of subterms |
+| `lift_eq` | `tt` | | Prove reflexive relations using proofs of equality (?) |
+| `use_axioms` | `tt` | | Allow simplifications that require `propext` or `funext` (?) |
```
digama0, line 413:
```suggestion
| `use_axioms` | `tt` | | Allow using `propext` and `funext` to rewrite under binders and use `A ↔ B` `simp` lemmas |
```

templates/extras/simp.md
kmill: Thanks for catching that

templates/extras/simp.md
```diff
@@ -62,222 +100,346 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
+```lean
+structure foo := (n : ℕ)
+
+def my_foo : foo := {n := 37}
+```
+then if you add a `simp` lemma that `my_foo.n = 37`, you give the simplifier the
+ability to evaluate the `foo.n` projection for `my_foo`, which saves you from having
+to unfold the definition of `my_foo` (by default `simp` does not unfold most definitions).
+Creating these `simp` lemmas is so common that there is an attribute that creates
+them all for you automatically:
```
bryangingechen, line 162:
```suggestion
Creating these `simp` lemmas is so common that there is
[an attribute](https://leanprover-community.github.io/mathlib_docs/attributes.html#simps)
that creates them all for you automatically:
```

templates/extras/simp.md
```diff
@@ -62,222 +100,346 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
+```lean
+@[simp] lemma mem_coe {a : α} {s : finset α} : a ∈ (s : set α) ↔ a ∈ s := ...
+```
+to replace occurrences of `a ∈ (s : set α)` with the correct normal form.
+
+Because the simplifier works from the inside out, simplifying
+arguments of a function before simplifying the function, a `simp`
+lemma should have the arguments to the function on its left-hand side in simp-normal
+form. For example if `g 0` can be simplified, then `@[simp] lemma foo : f (g 0) = 0` will never be used.
+Mathlib's `simp_nf` linter checks for this
```
bryangingechen, line 247:
```suggestion
Mathlib's `simp_nf` [linter](https://leanprover-community.github.io/mathlib_docs/commands.html#linting%20commands) checks for this
```

templates/extras/simp.md
```diff
@@ -62,222 +100,346 @@ lemmas". Here are some examples of `simp` lemmas in mathlib:
...
-relevant for the average user.
-
+### Configuration options
+
+Both `simp` and `dsimp` can take additional configuration options using record syntax.
+For example, `simp {single_pass := tt}` runs `simp` with the `single_pass` configuration option set to true.
+One can use `single_pass` to avoid loops that might otherwise occur.
+
+The core Lean file `init/meta/simp_tactic.lean` reveals other configuration options in
+the `dsimp_config` and `simp_config` structures.
```
bryangingechen, line 389:
```suggestion
the [`dsimp_config`](https://leanprover-community.github.io/mathlib_docs/find/tactic.dsimp_config/src) and [`simp_config`](https://leanprover-community.github.io/mathlib_docs/find/tactic.simp_config/src) structures.
```

bryangingechen (approved):
LGTM, thanks for taking this on! 

I suggested some links; feel free to reject any of them if you think they're unnecessary / distracting.

PR #197, opened 2021-08-22
feat(templates/learn): add link to youtube channel

semorrison:


PatrickMassot:
The one item list is a bit weird. Do we want a full list of our playlists? Or remove that list?

PR #198, opened 2021-08-22
Update leanproject documentation

eric-wieser:
This describes the new get-cache behavior, and pushes the "global mathlib install" section right to the bottom.

~~Note that this reflects the behavior only in the presence of https://github.com/leanprover-community/mathlib-tools/pull/113; right now if you run `get-cache --rev` in a non-mathlib project it will revert changes to lean files too, which seems unhelpful.~~

### DIFF
--- a/templates/leanproject.md
+++ b/templates/leanproject.md
@@ -52,7 +52,7 @@ leanproject build
 
 ### Getting mathlib oleans
 
-In an existing project depending on mathlib, you can run:
+In an existing project *depending on* mathlib (but not mathlib itself), you can run:
 ```text
 leanproject get-mathlib-cache
 ```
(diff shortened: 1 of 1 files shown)


templates/leanproject.md
```diff
@@ -52,8 +52,8 @@ leanproject build
 
 ### Getting mathlib oleans
 
-In an existing project depending on mathlib (or in mathlib itself), you
-can run:
+In an existing project *depending on* mathlib (but not mathlib itself), you can
```
jcommelin, line 55:
```suggestion
In an existing project *depending on* mathlib
(for mathlib itself, use `leanproject get-cache`, see below), you can
```

PR #201, opened 2021-09-20
Add a bit of doc about removing imports

alexjbest:
I kept it short as the page is already a lot of information


PatrickMassot:
Thanks!

PR #202, opened 2021-09-23
add CICM papers and others to the bibliography

robertylewis:
Things got renormalized, so the diff is hard to read. Look at the second and third commits.

### DIFF
--- a/lean.bib
+++ b/lean.bib
@@ -254,6 +254,34 @@
   tags          = {about-lean, lean2}
 }
 
+@InProceedings{   deMouraUllrich2021,
+  author        = "Moura, Leonardo de and Ullrich, Sebastian",
+  editor        = "Platzer, Andr{\'e} and Sutcliffe, Geoff",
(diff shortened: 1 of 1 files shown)


lean.bib
```diff
@@ -373,6 +412,15 @@ @Misc{            Huisinga19
   tags          = {about-lean, lean4}
 }
 
+@Booklet{         Kudryashov2021,
+  author        = {Yury Kudryashov},
+  title         = {Formalizing Rotation Number and Its Properties in Lean},
+  howpublished  = {EasyChair Preprint no. 6168},
+  url           = {https://easychair.org/publications/preprint/38w1},
+  year          = {2021},
+  tags          = {about-mathlib, lean3}
```
bryangingechen: What's the criterion for the `about-mathlib` tag? Some of the other papers (e.g. Eric's) could have that tag as well, right?

robertylewis: This was a copy-paste mistake, I fixed it in the next commit. This one is clearly `formalization`. Eric's is maybe debatable.

PR #208, opened 2021-10-10
Add first entries to a Lean glossary

Julian:
This came out of an idea from a discussion with @yaeldillies to put together a glossary of some terminology used conversationally in Lean and in mathlib.

### DIFF
--- a/templates/glossary.md
+++ b/templates/glossary.md
@@ -131,12 +131,18 @@ Further detail can be found in the [Lean documentation](https://leanprover.githu
 
 ### diamond
 
-There are often many ways to turn a given structure into another one.
-A *diamond* is a collection of such ways.
-Diamonds are abundant because of [hierarchies](#hierarchy).
-[Typeclass inference](#typeclass-inference) will unpredictably take one of the paths for a given diamond, so we want this path to not matter.
-This amounts to making the [Type-valued](#Prop-vs-Type) fields of the different inferable structures [defeq](#defeq).
-When this is the case, we have a *defeq diamond*.
+The existence of multiple conflicting [terms](#term) of a [class](#class) found within the typeclass [instance](#instance) graph.
+A diamond is likely to cause issues during [typeclass inference](#typeclass-inference), which attempts to construct a single term of the class and may therefore be unable to do so.
+When unqualified, "diamond" refers most often to this undesirable case, where non-[defeq](#defeq) terms in the diamond may cause errors, lead to [goals](#goal) which are not provable via `refl`, or potentially ones not provably equal at all.
+Within [mathlib](#mathlib), diamonds are abundant because of its many [hierarchies](#hierarchy).
+Fixing or mitigating diamonds often involves refactoring the fields or instance priorities for the offending class.
+Diamonds which cross library boundaries -- such as ones in which part of the typeclass graph lives within mathlib and part within a library depending on mathlib which adds new instances or classes -- may be particularly hard to fix or avoid without modification.
+
+#### See Also
+
+* [mathlib's Design note on `add_monoid` and `monoid`](https://leanprover-community.github.io/mathlib_docs/algebra/group/defs.html#design-note-on-codeadd_monoidcode-and-codemonoidcode), for a specific example of evading a diamond
+
+* [Forgetful Inheritance](https://leanprover-community.github.io/mathlib_docs/notes.html#forgetful%20inheritance), also from the mathlib documentation, for a discussion on a general pattern for avoiding diamonds in the case of "richer" and poorer structures on a type
 
 ### `equiv`
 


templates/glossary.md
```diff
@@ -0,0 +1,145 @@
...
+
+* [Section 5 of Theorem Proving in Lean](https://leanprover.github.io/theorem_proving_in_lean/tactics.html)
+
+* [The mathlib tactics documentation](https://leanprover-community.github.io/mathlib_docs/tactics.html)
+
+### term
+
+### term mode
+
+### terminal (or non-terminal) `simp`
```
jcommelin, line 139: I think this should get a cross-ref from an entry at `` `non-terminal `simp` ``.

jcommelin (commented):
Great idea. It would be good to crowd source this!

templates/glossary.md
```diff
@@ -0,0 +1,253 @@
...
+
+### class
+
+A *class* (aka *typeclass*) is a [structure](#structure) that can be handled by [typeclass inference](#typeclass-inference).
+
+### coercion / ↑
+
+### code linter
+
+A *code linter* is a [linter](#lint) concerned with how code works. Concretely, this is [a collection of Lean programs](https://leanprover-community.github.io/mathlib_docs/tactic/lint/frontend.html) checking that... Errors are flagged in the [nolint file](https://github.com/leanprover-community/mathlib/blob/master/scripts/nolints.txt).
```
jcommelin, line 67:
```suggestion
A *code linter* is a [linter](#lint) concerned with how code works. Concretely, this is [a collection of Lean programs](https://leanprover-community.github.io/mathlib_docs/tactic/lint/frontend.html) checking that.
[TODO]
Errors are flagged in the [nolint file](https://github.com/leanprover-community/mathlib/blob/master/scripts/nolints.txt).
```
Do I understand correctly that `...` indicates that this was unfinished? If so, I suggested inserting the word `TODO`, so that we can quickly spot such places.

templates/glossary.md
```diff
@@ -0,0 +1,253 @@
...
+
+Some `refl` invokations take an obnoxiously long time. There can be many reasons for this. See [this Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/refl.20taking.2020.20seconds).
+
+### hierarchy
+
+A *hierarchy* is a collection of [typeclasses](#typeclass) which are more and more constraining. In mathlib, we have the *algebraic hierarchy* (`semiring`, `ring`, `field`, ...), the *order hierarchy* (`preorder`, `partial_order`, `linear_order`, ...), the *topological hierarchy* (`t1_space`, `t2_space`, `normal_space`, ...), the *categorical hierarchy* (`preadditive`, `abelian`, `monoidal`, ...) but also the *scalar hierarchy* (`mul_action`, `distrib_mul_action`, `module`, ...), the *norm hierarchy*, and intersection of previous ones like the *order-algebraic hierarchy*, the *topologico-algebraic hierarchy*, ...
+
+### `Icc`, `Ico`, `Ioc`, `Ioo`, `Ici`, `Ioi`, `Iic`, `Iio`
+
+There are `9` types of intervals, depending on whether the interval is **c**losed, **o**pen or runs to **i**nfinity at each end. Names have been made compact by calling each interval `I` + how it ends on the left + how it ends on the right.
```
jcommelin, line 122: I think `Iii` is not really a thing right? So effectively there are only 8.

templates/glossary.md
```diff
@@ -0,0 +1,253 @@
...
+
+### lift
+
+### lint
+
+A *linter* is a small program that looks for hard-to-spot mistakes in code. For mathlib, we use [style linters](#style-linter)* and [code linters](#code-linter). Mathlib is linted at every [CI run](#continuous-integration) after being built.
+
+### locale
+
+A *locale* is akin to a mathematical environment.
```
jcommelin, line 157: This could use a bit more explanation.

templates/glossary.md
```diff
@@ -0,0 +1,253 @@
...
+
+
+
+### core Lean
+
+The part of Lean written in Lean itself. mathlib was originally an annex of Lean. Since then, it outgrew it. But some basic lemmas and definitions are still held in *core Lean*, [the Lean repository](https://github.com/leanprover-community/lean), which renders it much less flexible.
+
+### declaration
+
+A *declaration* is anything that starts with `def`, `lemma`, `theorem`. This corresponds to creating an (or several) object in the background environment.
```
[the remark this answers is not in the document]
Julian: Is it not both the syntactical command itself (of which these are 3 examples but [the lean4 parser has a bunch more](https://github.com/leanprover/lean4/blob/master/src/Lean/Parser/Command.lean#L85-L86)) as well as the resulting object or objects it ends up declaring?

I see the lean4 doc page doesn't have a heading written yet [but does at least refer to these as basic declarations](https://leanprover.github.io/lean4/doc/declarations.html#basic-declarations), and seemingly is referring to the syntactical element a bit more than the resulting object, though it's mixing a bit I suppose.

templates/glossary.md
```diff
@@ -0,0 +1,253 @@
...
+
+### concrete structure
+
+### continuous integration
+
+
+
+### core Lean
+
+The part of Lean written in Lean itself. mathlib was originally an annex of Lean. Since then, it outgrew it. But some basic lemmas and definitions are still held in *core Lean*, [the Lean repository](https://github.com/leanprover-community/lean), which renders it much less flexible.
```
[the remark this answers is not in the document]
Julian: @YaelDillies I reworded this a bit, hopefully without bastardizing what you wrote too badly, and addressing what Rob mentioned.

templates/glossary.md
```diff
@@ -0,0 +1,253 @@
+# Lean Glossary
+
+### abbreviation
```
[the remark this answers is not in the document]
Julian: (I didn't do this zulip searching yet, though IIRC the `abbreviation` statement is gone from Lean 4, there's just `abbrev` now I think). But yes still need to do this.

templates/glossary.md
```diff
@@ -159,11 +159,11 @@ Names have been made compact by calling each interval `I` + how it ends on the l
 ### instance
 
 "instance" refers to two closely related concepts:
+
 * An *instance* is a [class](#class) argument to a `def`/`lemma`.
 They are put in square brackets `[]` for [typeclass inference](#typeclass-inference) to pick them up when processing the statement.
 * An *instance* is an `instance` [declaration](#declaration). Instances For example, `ℝ` has a linear
 
-
 ### instance vs def
```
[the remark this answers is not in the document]
jcommelin: Sounds good.

jcommelin:
Personally, I think it's fine to split this PR. A glossary is never finished. So if there's a bunch of entries that are "stable", we could split them into a new PR and have them go live. And then new entries can be added in small batches.
Otherwise we run the risk that this will be such a large PR that there will always be something that is unfinished or insufficiently bikeshedded.

templates/glossary.md
```diff
@@ -0,0 +1,272 @@
...
+### `equiv`
+
+As distinct from mathematical equality, [`equiv`](mathlib_docs/data/equiv/basic.html) allows for defining an equivalence or congruence of types.
+One important thing to note is that an `equiv` [holds data instead of being merely a proof](#bundled-vs-unbundled).
+
+### goal
+
+Within the context of interactively proving theorems in Lean, each targeted statement whose proof is in-progress.
+
+Or more broadly, type theoretically, the process of exhibiting a term of a specific desired type.
```
robertylewis, line 109: This doesn't really match -- "goal" isn't a process.

templates/glossary.md
```diff
@@ -0,0 +1,272 @@
...
+[This Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/refl.20taking.2020.20seconds) has a particularly slow example.
+
+### hierarchy
+
+A collection of successively more constrained [typeclasses](#typeclass) within a related area of mathematics.
+In [mathlib](#mathlib), we have the *algebraic hierarchy* (`semiring`, `ring`, `field`, ...), the *order hierarchy* (`preorder`, `partial_order`, `linear_order`, ...), the *topological hierarchy* (`t1_space`, `t2_space`, `normal_space`, ...), the *categorical hierarchy* (`preadditive`, `abelian`, `monoidal`, ...) but also the *scalar hierarchy* (`mul_action`, `distrib_mul_action`, `module`, ...), the *norm hierarchy*, and intersections of previous ones like the *order-algebraic hierarchy*, the *topologico-algebraic hierarchy*, and others.
+
+### `Icc`, `Ico`, `Ioc`, `Ioo`, `Ici`, `Ioi`, `Iic`, `Iio`
+
+There are `9` types of intervals, depending on whether the interval is **c**losed, **o**pen or runs to **i**nfinity at each end.
```
robertylewis, line 126:
```suggestion
There are `8` types of intervals, depending on whether the interval is **c**losed, **o**pen or runs to **i**nfinity at each end.
```
Worth mentioning that `Iii` isn't a thing? I'm fine either way.

templates/glossary.md
Julian: Slightly reworded. It's still a bit clunky but was the best I had for the minute, let me know how it looks now to you.

templates/glossary.md
Julian: Done.

templates/glossary.md
robertylewis: Looks good to me!

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
+# Lean Glossary
+
```
robertylewis, line 2: Would it make sense to add a sentence or two about what this document is? Maybe note that these entries aren't intended to be definitions, but to clarify their typical usage in the Lean community/Zulip. I say this because a lot of the terms do have technical meanings that aren't precisely captured here.

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+### bundled vs unbundled
+
+Given a mathematical object `O` with property `P`, bundling `P` refers to the creation of a Lean [structure](#structure) containing a proof of `P` as one of its fields in addition to those needed to structurally define `O`.
+
+In contrast, an unbundled structure instead contains only the definition of `O`, and separately the creation of an `is_P` proposition which can be applied to terms of `O`.
+
+As a concrete example, a [group homomorphism](https://en.wikipedia.org/wiki/Group_homomorphism) can be viewed as a map `φ: G → H` between groups, along with a proof `h : φ(a * b) = φ(a) * φ(b)`. A bundled group homomorphism would contain both `φ` and `h` as fields, whereas an unbundled one would instead consist only of `φ`, with a separate `is_group_homomorphism` [declaration](#declaration) for the proof of `h`.
+
+There are performance, stylistic or implementation-related reasons to prefer bundling or unbundling.
+Generally within [mathlib](#mathlib), bundled structures are preferred, but unbundled versions are often also present.
```
robertylewis, line 46: I'm unsure about this line, which doesn't exactly contradict 4.1 of https://arxiv.org/pdf/1910.09336.pdf , but also doesn't exactly agree. It might be good to reference this section of the paper here too.

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+A [mode](#mode) which consists of sequences of successive transformations of expressions involving a transitive relation such as `=`, `<` or others tagged with `trans` [attributes](#attribute).
+It is entered via the `calc` keyword.
+
+#### See also
+
+[The `calc` mode community documentation](https://leanprover-community.github.io/extras/calc.html)
+
+### carrier
+
+For a [bundled](#bundled-vs-unbundled) mathematical subobject on a type `T`, it is the *set* of [terms of `T`](#term) considered within the subobject.
```
robertylewis, line 67: I don't think "carrier" has any particular connection to subobjects, does it? If we used fully bundled groups, the type of elements would (could) be called `carrier`. It would make mathematical sense to call `T` in `[group T]` the carrier type. https://en.wikipedia.org/wiki/Algebraic_structure

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+### goal
+
+Within the context of interactively proving theorems in Lean, each targeted statement whose proof is in-progress.
+
+Or more broadly, type theoretically, an individual type for which a [term](#term) is to be exhibited.
+For propositions, exhibiting a [term](#term) reduces equivalently to the aforementioned idea of proof.
+
+### golfing
+
+An attempt or effort to make a particular functional piece of code as short as possible; within the context of Lean, often it is a proof's length which is golfed or optimized.
```
robertylewis, line 133:
```suggestion
An attempt or effort to make a particular working piece of code as short as possible; within the context of Lean, often it is a proof's length which is golfed or optimized.
```
to avoid others experiencing my half-second of confusion here? :smile:

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+Golfed proofs often make use of [term mode](#term-mode), and in general prioritize terseness over readability.
+This obfuscation often is used intentionally or advantageously to signal to readers that a proof is mechanical or trivial.
+
+### heavy `rfl` / heavy `refl`
+
+A use of `rfl` (or `refl`) which performs slowly when it is evaluated by Lean.
+A heavy `rfl` often occurs when `rfl` is asked to perform too many steps of definitional reduction at once, resulting in a proof term which is small but which requires Lean to do a lot of computation to ensure that it type checks.
```
robertylewis, line 140:
```suggestion
A heavy `rfl` occurs when `rfl` is asked to perform many steps of definitional reduction at once, resulting in a proof term which is small but which requires Lean to do a lot of computation to ensure that it type checks.
```
Remove some hedging and a value judgment. Kernel computation isn't always evil!

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+Previous talks or presentations from the events have been [shared on the community YouTube channel](https://www.youtube.com/channel/UCWe5B7Ikr0AI9727doEUxPg).
+### `leanproject`
+
+A higher-level supporting tool for working with projects, particularly [mathlib](#mathlib), in Lean 3.
+It lives within the [community `mathlib-tools` repository](https://github.com/leanprover-community/mathlib-tools/).
+
+### lint
+
+A *linter* is a small program that looks for hard-to-spot mistakes in code.
+For [mathlib](#mathlib), we use [style linters](#style-linter)* and [code linters](#code-linter).
```
robertylewis, line 181: I don't see a code linter entry, but I thought there was one before? 

Is the * after the style linter link Markdown for something?

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+
+Lean code has to be compiled.
+The compiled version of file `x.lean` is file `x.olean` and all olean files together form the [cache](#cache).
+
+### orange bar of hell
+
+In VScode, the *orange bar of hell* refers to the orange bar that appears left to the current file when it persists.
+The reason is that the Lean extension has to (re)compile all the imported files whose [cache](#cache) does not match.
+Because of this, having branches that touch two files far apart (with respect to the [import tree](#import-tree)) related through imports is considered bad practice as **any** modification on the file upstream will force Lean to recompile all the files in the middle.
+
```
robertylewis, line 245: How about adding suggestions here for escaping hell? I give my students this snippet which could be adapted:

Lean in VSCode is usually pretty responsive. If it's feeling slow, or you're seeing orange lines on the side that aren't going away, there are a few things to try.

First, in VSCode, close any editor tabs that you aren't using anymore. Then open the Command Palette (ctrl-shift-p or cmd-shift-p) and run Lean: Restart. You might find it useful to map this to a hotkey in VSCode. I use ctrl-r.

If that doesn't work, let's make sure you have a fresh copy of the library. In the root fpv2021 directory, run:

```
leanpkg configure
leanproject get-mathlib
lean --make src/lovelib.lean
```

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+## propeq
+
+*Propositional equality*. Two [term](#term)s `a b : α` are propositionally equal if we can prove `a = b`.
+This is weaker than [definitional](#defeq) and [syntactical](#syntactical-equality) equalities.
```
robertylewis, line 249: Broken links afaict

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+### style linter
+
+A *style linter* is a [linter](#lint) concerned with how code looks.
+Concretely, this is [a short Python program](https://github.com/leanprover-community/mathlib/blob/master/scripts/lint-style.py) checking that all lines are less than `100` characters long, that every file has a [module docstring](#module-docstring)... Errors are flagged in the [style exceptions file](https://github.com/leanprover-community/mathlib/blob/master/scripts/style-exceptions.txt).
+
+### tactic mode
+
+A Lean [mode](#mode) characterized by its reliance on sequences of [tactics](#tactic) which often facilitate proofs quite similar to paper-based reasoning, albeit often with the use of sophisticated tactics which automate tedious portions of a proof.
+There are various means to [enter tactic mode](https://leanprover.github.io/theorem_proving_in_lean/tactics.html#entering-tactic-mode).
+It may be entered using the `by` keyword from [term mode](#term-mode), though in Lean 3 it is most often entered via a `begin...end` block whenever its body is made up of multiple commands.
```
robertylewis, line 278:
```suggestion
In Lean 3 it is most often entered via a `begin...end` block. The `by` keyword can also be used to switch from [term mode](#term-mode) to tactic mode for a single tactic.
```
It feels strange to me to describe the most common mode second.

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+Other modes can also be interspersed within it, often to collaboratively produce an understandable, efficient, short or readable overall proof.
+Ultimately, the result of a tactic mode block is a [term](#term), assembled via the tactics within it.
```
robertylewis, line 280: broken link

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+
+* [Section 5 of Theorem Proving in Lean](https://leanprover.github.io/theorem_proving_in_lean/tactics.html), which discusses tactics, as well as moving into and out of tactic mode
+
+* [The `show_term` tactic](https://leanprover-community.github.io/mathlib_docs/tactic/show_term.html) which can reveal the assembled [term](#term)
+
+* [The mathlib tactics documentation](https://leanprover-community.github.io/mathlib_docs/tactics.html), with a comprehensive list of tactics
+
+### term mode
+
+A Lean [mode](#mode) which assembles a single [term](#term) through the use of functional subexpressions.
```
robertylewis, line 292: term is broken link

templates/glossary.md
```diff
@@ -0,0 +1,331 @@
...
+In contrast to [tactic mode](#tactic-mode), term mode proofs are often short in length, though potentially harder to read for humans.
+There are various means to enter term mode.
+The body of a [declaration](#declaration) begins in term mode, or within [tactic mode](#tactic-mode) it is often entered using the `exact` [tactic](#tactic).
+Efficient term mode proofs often contribute to [code golfing](#golfing).
```
robertylewis, line 296:
```suggestion
Efficient term mode proofs often contribute to [code golfing](#golfing).

Commands such as `have`, `suffices`, and `show` can be used to write structured term mode proofs that are more legible than bare proof terms.
```

robertylewis:
It would make sense to add a link to the glossary from the sidebar as part of the same PR.

templates/glossary.md
Julian: Yes definitely a good idea, done.

templates/glossary.md
robertylewis: I don't feel strongly, either is fine.

templates/glossary.md
Julian: Ha, indeed.

templates/glossary.md
Julian: Ok, will leave as-is then, thanks!

templates/glossary.md
Julian: Good idea.

templates/glossary.md
Julian: Done, thanks!

templates/glossary.md
```diff
@@ -0,0 +1,356 @@
...
+
+The `@[simp]` attribute, for example, tags an object (ostensibly a `lemma` or `theorem`) as being a [simp lemma](#simp-lemma).
+
+For further details, including a list of commonly used attributes, see [this section of the Lean reference manual](https://leanprover.github.io/reference/other_commands.html#attributes).
+
+### beta reduction
+
+A specific simplification operation in [dependent type theory](#dependent-type-theory) (and Lean's implementation of it) which may be performed as part of deciding whether two [terms](#term) are [definitionally equal](#defeq).
+
+More precisely, it is the process of simplifying an expression such as `(λ x, x) a` by replacing appearances of the variable `x` with `a`.
```
robertylewis, line 23:
```suggestion
More precisely, it is the process of simplifying an expression such as `(λ x, t) a` to `t[a/x]`, where appearances of the variable `x` in `t` have been replaced with `a`.
```

templates/glossary.md
```diff
@@ -0,0 +1,356 @@
...
+
+Distinct from the use of *class* in object oriented languages -- the word usage in functional programming languages comes from the [typeclasses of Haskell](https://en.wikipedia.org/wiki/Type_class).
+
+### code linter
+
+A *code linter* is a [linter](#lint) which attempts to find unintentional mistakes in Lean code within [mathlib](#mathlib).
+Concretely, mathlib includes [a collection of Lean programs](https://leanprover-community.github.io/mathlib_docs/tactic/lint/frontend.html) which check for a variety of potential issues which may be introduced in new mathlib code.
+New linters are occasionally introduced to detect additional classes of mistakes.
+Mathlib's [continuous integration](#continuous-integration) ensures that any new such code passes the defined linters.
+Linting can be disabled for a particular piece of code, which will be indicated in an autogenerated [nolint file](https://github.com/leanprover-community/mathlib/blob/master/scripts/nolints.txt).
```
robertylewis, line 94:
```suggestion
Linting can be disabled for a particular piece of code by using the `nolint` [attribute](#attribute). Some linter failures that predate the CI code linters are indicated in an autogenerated [nolint file](https://github.com/leanprover-community/mathlib/blob/master/scripts/nolints.txt); the length of this file should tend to zero over time as these errors are fixed.
```

Julian:
Thanks, applied, both of those made sense to me as well, appreciated.

PR #211, opened 2021-10-15
Update windows.md

c0ppelius:
Updated with some instructions for installing Lean, Mathlib, and VS Code through WSL.


bryangingechen:
@pechersky do you mind taking a look at these instructions too?

c0ppelius:
No need to apologize for the delay given I did even worse! I added a few lines for justification. Mainly: use this if you are comfortable in a linux style terminal and/or as a fallback install method in case Windows is giving you fits.

PR #218, opened 2021-11-01
Add a link to the import graph

eric-wieser:
Putting this up for discussion. We could also consider embedding the graph, but then we'd need some way to strip the UI elements that won't scale well.

We might want this under the `leanprover-community` org if we plan to embed it.


bryangingechen (approved):
LGTM, but others may have stronger opinions.

robertylewis:
This has sat needlessly for too long and I've gone looking for this link before!

PR #222, opened 2021-11-04
Update style.md

hrldcpr:
I noticed that this page doesn't always follow its own “Use a space after binders” rule, so went ahead and added some spaces


bryangingechen:
Nice catch, thanks!

PR #223, opened 2021-11-04
Clarify which VSCode extension to install

JLimperg:
Some of the installation instructions used to tell people to search for
the 'leanprover' VSCode extension. This is ambiguous since it can refer
to both 'lean' and 'lean4' (and 'lean4' seems to be listed first now).

This commit clarifies that 'lean' is the correct extension.


bryangingechen:
Thanks!

Issue #226, opened 2021-11-22
Add a glossary entries for HoTT, UH, TC, fbip, dot notation and API


gabriel-doriath-dohler:
These terms and acronyms could be difficult for newcomers depending on their background:
- HoTT: Homotopy Type Theory
- UH: Unification Hints
- TC: Type Classes
- fbip: Functional But In Place
- dot notation: field notation: generalized structure projection
- API: Application Programming Interface

Julian:
(#249 gave us HoTT, so we can cross that one off).

I'll hit API at some point soon.

PR #227, opened 2021-11-22
feat(templates/glossary): explain dot notation

Vierkantor:
As mentioned in #226.

### DIFF
--- a/templates/glossary.md
+++ b/templates/glossary.md
@@ -155,7 +155,9 @@ Diamonds which cross library boundaries -- such as ones in which part of the typ
 
 The syntax sugar allowing notations such as `((foo a b c).bar x.y).baz`.
 
-Lean provides two interpretations of the syntax `foo.bar`: it can mean the declaration `bar` in the `foo` namespace, or it can be generalized field notation. Suppose `foo` has type `C x1 ... xn`, with `C` some constant and `x1 ... xn` arbitrary, then `foo.bar` is sugar for `C.bar foo`. For calls of the form `foo.bar _ ... _` with (implicit or explicit) arguments, Lean is smart enough to expand to `C.bar _ ... foo _ ... _`, so that everything typechecks. In these previous examples, `foo` can also be a more complicated expression such as function application.
+Lean provides two interpretations of the syntax `foo.bar`: it can mean the declaration `bar` in the `foo` namespace, or it can be generalized field notation. We expand here on the latter. 
+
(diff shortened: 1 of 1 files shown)


templates/glossary.md
```diff
@@ -149,6 +149,12 @@ Diamonds which cross library boundaries -- such as ones in which part of the typ
 
 * [Forgetful Inheritance](https://leanprover-community.github.io/mathlib_docs/notes.html#forgetful%20inheritance), also from the mathlib documentation, for a discussion on a general pattern for avoiding diamonds in the case of "richer" and poorer structures on a type
 
+### dot notation / generalized field notation / generalized projections
```
Julian, line 152:
```suggestion
<a name="dot-notation"></a>

### dot notation / generalized field notation / generalized projections
```

will make it easier to link this from Zulip (as `glossary#dot-notation`).

Julian (approved):
Yay!

templates/glossary.md
```diff
@@ -149,6 +149,14 @@ Diamonds which cross library boundaries -- such as ones in which part of the typ
 
 * [Forgetful Inheritance](https://leanprover-community.github.io/mathlib_docs/notes.html#forgetful%20inheritance), also from the mathlib documentation, for a discussion on a general pattern for avoiding diamonds in the case of "richer" and poorer structures on a type
 
+<a name="dot-notation"></a>
+
+### dot notation / generalized field notation / generalized projections
+
+The syntax sugar allowing notations such as `((foo a b c).bar x.y).baz`.
+
+Lean provides two interpretations of the syntax `foo.bar`: it can mean the declaration `bar` in the `foo` namespace, or it can be generalized field notation. Suppose `foo` has type `C x1 ... xn`, with `C` some constant and `x1 ... xn` arbitrary, then `foo.bar` is sugar for `C.bar foo`. For calls of the form `foo.bar _ ... _` with (implicit or explicit) arguments, Lean is smart enough to expand to `C.bar _ ... foo _ ... _`, so that everything typechecks. In these previous examples, `foo` can also be a more complicated expression such as function application.
```
robertylewis, line 158:
```suggestion
Lean provides two interpretations of the syntax `foo.bar`: it can mean the declaration `bar` in the `foo` namespace, or it can be generalized field notation. We expand here on the latter. 

Suppose `foo` has type `C x1 ... xn`, with `C` some constant and `x1 ... xn` arbitrary, and suppose that there is a declaration in the context named `C.bar` which takes an argument of type `C x1 ... xn`. Then `foo.bar` is sugar for `C.bar foo`. For calls of the form `foo.bar _ ... _` with (implicit or explicit) arguments, Lean is smart enough to expand to `C.bar _ ... foo _ ... _`, so that everything typechecks. In these previous examples, `foo` can also be a more complicated expression such as function application.
```

Tried to clarify slightly, since superficially it looks like we're defining dot notation in terms of dot notation.

I'm not sure what you mean by the final sentence "In these previous examples, `foo` can also be a more complicated expression such as function application."?

Julian: I think @Vierkantor was pointing out that `(foo bar baz).quux` also works, which I think is not a bad thing to point out; maybe just giving that full example will make it clearer.

robertylewis: @Vierkantor do you like these changes, and do you think @Julian 's suggestion to add an example of a "more complicated expression" would fit?

PR #230, opened 2021-12-07
Adding text on "orange bars of hell"

stuart-presnell:
As discussed in this Zulip thread (https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Troubleshooting.20mathlib), this page is the only relevant result that's returned by a Google search for "troubleshooting mathlib", so it seems worthwhile to put some more general troubleshooting help here.

### DIFF
--- a/templates/leanproject.md
+++ b/templates/leanproject.md
@@ -253,3 +253,27 @@ If `leanproject` ends with a mysterious error message, you can run it
 using the `--debug` flag, e.g. `leanproject --debug new my_project`.
 It will then probably output a python trace that you'll be able to paste
 in a GitHub issue or on [Zulip](https://leanprover.zulipchat.com/).
+
+Another common problem you may encounter when editing mathlib files is the 
+["orange bars of hell"](https://leanprover-community.github.io/glossary.html#orange-bar-of-hell).
+When this problem arises the orange bars in the sidebar of the editor 
+(e.g. VSCode) alongside a Lean file, which normally indicate which parts of 
+the file Lean is still evaluating, refuse to disappear or update.  
+
+* The first thing to try if this occurs is to close any inactive editor 
+tabs and then restart Lean. (In VSCode, press `ctrl-shift-p` or `cmd-shift-p` 
+to open the Command Palette and then run the command `Lean: Restart`). 
+
+* If the problem persists after restarting, this may be because your local 
+copy of the 
+(pre-compiled olean files)[https://leanprover-community.github.io/glossary.html#cache] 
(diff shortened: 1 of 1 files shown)


templates/leanproject.md
```diff
@@ -253,3 +253,27 @@ If `leanproject` ends with a mysterious error message, you can run it
...
+
+* The first thing to try if this occurs is to close any inactive editor 
+tabs and then restart Lean. (In VSCode, press `ctrl-shift-p` or `cmd-shift-p` 
+to open the Command Palette and then run the command `Lean: Restart`). 
+
+* If the problem persists after restarting, this may be because your local 
+copy of the 
+(pre-compiled olean files)[https://leanprover-community.github.io/glossary.html#cache] 
+is mismatched or faulty in some other way, and so Lean is having to (re)compile 
+all the imported mathlib files.  This taks a long time, making Lean appear
```
PatrickMassot, line 271:
```suggestion
all the imported mathlib files.  This takes a long time, making Lean appear 
```

templates/leanproject.md
```diff
@@ -253,3 +253,27 @@ If `leanproject` ends with a mysterious error message, you can run it
...
+to open the Command Palette and then run the command `Lean: Restart`). 
+
+* If the problem persists after restarting, this may be because your local 
+copy of the 
+(pre-compiled olean files)[https://leanprover-community.github.io/glossary.html#cache] 
+is mismatched or faulty in some other way, and so Lean is having to (re)compile 
+all the imported mathlib files.  This takes a long time, making Lean appear 
+unresponsive.  In this case, make sure any changes in your current branch 
+are committed, switch to the `master` branch, and then re-download the mathlib 
+cache by typing `leanpkg configure && leanproject get-mathlib-cache` at the terminal.
```
robertylewis, line 274: I think this is confused about whether you're working on mathlib or a project that depends on mathlib. If it's a dependent project, then switching to master won't do anything. But these problems certainly do show up in external projects, especially if you accidentally touch a file in `_target`.

stuart-presnell: Good point, I was only thinking in the context of working on mathlib. Should we have different advice here for the two different cases, or a parenthetical note?

stuart-presnell: Or is what I've written just incorrect in either case?

eric-wieser: The part about switching to master is bad advice here, as I think I mentioned in a recent zulip thread.

semorrison:
I think this is no longer relevant now that we're switching to Lean 4. Apologies in any case for letting a PR languish for so long!

PR #232, opened 2021-12-15
Recategorize the scalar actions paper as about mathlib

eric-wieser:


eric-wieser:
Ah, I now see this was discussed briefly [here](https://github.com/leanprover-community/leanprover-community.github.io/pull/202#discussion_r715097568). Would you prefer to leave this where it is?

robertylewis:
Up to you, I think it fits in both buckets equally well!

PatrickMassot:
Since it fits both, let's let the author decide.

PR #234, opened 2021-12-21
Add glossary entries for {dependent ,}type theory

Julian:
I tried to mix a bit of inline explanation with some links, but obviously this is not my area of expertise so as usual comments are welcome on whether the right balance is struck (or whether anything's completely wrong of course).

(there's also a couple of lines of diff noise fixing some minor interlinks and whitespace, sorry for mixing that in)


templates/glossary.md
```diff
@@ -134,9 +137,27 @@ Examples of such commands are the `def`, `theorem`, `constant` or `example` comm
 
 Further detail can be found in the [Lean documentation](https://leanprover.github.io/lean4/doc/declarations.html#basic-declarations).
 
+### dependent type theory
+
+A [type theory](#type-theory) in which one additionally may have types which depend on a parameter, such as the type of "days of a calendar month", whose specific values depend on the particular month, given that different months have different numbers of days.
+Further examples may be found in the resources below.
+Lean's implementation of dependent type theory is based on what is known as the *Calculus of Constructions*, allowing for its use both in complex mathematical reasoning as well as software verification.
```
Julian, line 144: Here I was going to add some additional notes but I think if we have entries for pi type and sigma type (which probably we should) that the right thing to do will be to then just mention those two here, but thoughts welcome.

templates/glossary.md
```diff
@@ -351,7 +373,20 @@ Commands such as `have`, `suffices`, and `show` can be used to write structured
...
+"[Theorem Proving in Lean](https://leanprover.github.io/theorem_proving_in_lean/index.html)", a free online textbook by Jeremy Avigad, Leonardo de Moura, and Soonho Kong, "designed to teach you to develop and verify proofs in Lean".
+Starting with a simple introduction to the [type theory](https://leanprover.github.io/theorem_proving_in_lean/dependent_type_theory.html) used in Lean, the text proceeds to explain topics such as [tactics](https://leanprover.github.io/theorem_proving_in_lean/tactics.html), [inductive types](https://leanprover.github.io/theorem_proving_in_lean/inductive_types.html), and [type classes](https://leanprover.github.io/theorem_proving_in_lean/type_classes.html).
+
+### type theory
+
+A formal system with two fundamental kinds of objects, [terms](#term) and types.
+It also may refer to the field of study of such languages, as there are nuances between various additional properties or features a specific type theory may contain.
+
+Lean's [dependent](#dependent-type-theory) type theory underlies its foundational system of mathematics in contrast to set theory.
```
Julian, line 384: I'm not 100% happy with getting to the definition here via having to start with set theory before getting to the type theory definition but I couldn't think of a better way to describe this understandably to someone otherwise (certainly it didn't really click for me personally until seeing something like the below example). Ideas welcome for how to reorder though if someone has one.

Julian:
Polite ping for any comments.

templates/glossary.md
```diff
@@ -134,9 +137,27 @@ Examples of such commands are the `def`, `theorem`, `constant` or `example` comm
 
 Further detail can be found in the [Lean documentation](https://leanprover.github.io/lean4/doc/declarations.html#basic-declarations).
 
+### dependent type theory
+
+A [type theory](#type-theory) in which one additionally may have types which depend on a parameter, such as the type of "days of a calendar month", whose specific values depend on the particular month, given that different months have different numbers of days.
```
robertylewis, line 142: Do you find this example clearer than `vector A n`, the type of `n`-tuples of elements of `A`? I'm not reading it as a novice but I feel like I stumbled over the days example more than I would have.

templates/glossary.md
```diff
@@ -134,9 +137,27 @@ Examples of such commands are the `def`, `theorem`, `constant` or `example` comm
...
+* [Dependent Type Theory from nLab](https://ncatlab.org/nlab/show/dependent+type+theory), for a general treatment of dependent type theories
+
+* [Mike Shulman's In Praise of Dependent Types](https://golem.ph.utexas.edu/category/2010/03/in_praise_of_dependent_types.html)
+
+* [Andrej Bauer's answer to "What makes dependent type theory more suitable than set theory for proof assistants?"](https://mathoverflow.net/a/376973)
+
 ### diamond
 
-The existence of multiple conflicting [terms](#term) of a [class](#class) found within the typeclass [instance](#instance) graph.
+The existence of multiple conflicting [terms](#term) of a [class](#class) found within the [typeclass](#class) [instance](#instance) graph.
```
robertylewis, line 160: `#class` is already linked in this sentence and with this change there are two different links right next to each other, which can look confusing -- our style sheet doesn't underline links. Do you think this is necessary?

templates/glossary.md
```diff
@@ -351,7 +373,20 @@ Commands such as `have`, `suffices`, and `show` can be used to write structured
...
+Starting with a simple introduction to the [type theory](https://leanprover.github.io/theorem_proving_in_lean/dependent_type_theory.html) used in Lean, the text proceeds to explain topics such as [tactics](https://leanprover.github.io/theorem_proving_in_lean/tactics.html), [inductive types](https://leanprover.github.io/theorem_proving_in_lean/inductive_types.html), and [type classes](https://leanprover.github.io/theorem_proving_in_lean/type_classes.html).
+
+### type theory
+
+A formal system with two fundamental kinds of objects, [terms](#term) and types.
+It also may refer to the field of study of such languages, as there are nuances between various additional properties or features a specific type theory may contain.
+
+Lean's [dependent](#dependent-type-theory) type theory underlies its foundational system of mathematics in contrast to set theory.
+Foundationally in set theory, all objects of interest are in a formal sense built out of a single kind of primitive object -- sets, and one reasons about whether constructed sets are or are not members of other sets.
```
robertylewis, line 385:
```suggestion
Foundationally in set theory, all objects are in a formal sense built out of a single kind of primitive object, sets. One reasons about whether constructed sets are or are not members of other sets.
```

templates/glossary.md
robertylewis: I think it's a reasonable approach.

robertylewis (commented):
Looks good to me.

templates/glossary.md
Julian: Definitely a good point, reverting.

templates/glossary.md
Julian: I think for me it clicked probably only after seeing a few different examples, but I found days of the month *slightly* and only very slightly easier to understand.

I think the reason was purely that for `vector A n` you have to stress that the length is part of the type, which is both a bit longer to writem as well as possible to miss the importance of if you're of certain background. In my head I mapped `vector A n` to Python list (somehow glossing over the length-part-of-type bit being important), and said "yeah what's the problem?", whereas with days of the month it was clearer that "oh yeah that'd be annoying".

But it's a tricky entry for novices, who knows if I even fully understand it myself yet...

Julian:
Thanks for the comments!

robertylewis:
Thanks!

PR #240, opened 2022-01-06
Add Tips and Tricks page

alexjbest:
For tips to help development that don't quite fit anywhere else, as suggested on Zulip (https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Tips.20and.20tricks/near/265408881)

Suggestions/additions welcome! I just put something rough together based on some old Zulip posts I wrote so far.

### DIFF
--- a/templates/tips_and_tricks.md
+++ b/templates/tips_and_tricks.md
@@ -25,7 +25,7 @@ In VSCode this can be configured by opening the settings and searching for `Lean
 
 ### Old mode
 If you change a file early on in the import hierarchy and then attempt to work on a file that depends on it much later, then Lean will try to check all files in between to ensure they didn't break.
-This behaviour can cause the Lean server to become unresponsive, or simply slow down development when.
+This behaviour can cause the Lean server to become unresponsive, or simply slow down development.
 You can also call lean with the "--old" flag to help in these situations, though it does come with some caveats.
 In VSCode you can enable it by opening the VSCode settings and searching for "Lean: extra options" then add the flag `--old`.
 This causes lean not to recompile oleans for unedited files and just assume everything worked, so in this situation it is helpful to avoid recompilation.
--- a/templates/tips_and_tricks.md
+++ b/templates/tips_and_tricks.md
@@ -5,7 +5,7 @@ It is currently focussed around Lean 3.
 
 ### Commenting out large proofs using `sorry`
 
-Sometimes commenting out chunks of a proof is helpful, often a convenient way to systematically do disable some proofs steps (rather than commenting out subsections) is to temporarily skip a long subproof (of a subgoal which has been surrounded by braces), you can do this by prepending `sorry;`
+Sometimes skipping subportions of a proof is helpful (e.g. to disable a portion which is slow while working on the rest). The straightforward way to do so is to comment out the lines which constitute the subportion, but if a subgoal has been surrounded by braces, you can also temporarily skip it by prepending `sorry;`
 ```
 sorry; { rw blah,
   ...
--- a/templates/tips_and_tricks.md
+++ b/templates/tips_and_tricks.md
@@ -5,14 +5,14 @@ It is currently focussed around Lean 3.
 
 ### Commenting out large proofs using `sorry`
 
-Sometimes commenting out large chunks of a proof is helpful, while the default comment characters are `--` or `/- multi-line comment-/`, however often a more convenient way is to temporarily skip a long subproof (of a focussed goal surrounded by braces) by prepending `sorry;`
+Sometimes commenting out chunks of a proof is helpful, often a convenient way to systematically do disable some proofs steps (rather than commenting out subsections) is to temporarily skip a long subproof (of a subgoal which has been surrounded by braces), you can do this by prepending `sorry;`
 ```
 sorry; { rw blah,
   ...
   HUGE PROOF,
   ... }
 ```
-With this method only one line needs to be changed, syntax highlighting is usually preserved, and jumping to definition still works. It also has the advantage that it is extremely easy to remove all instances of `sorry;` from a file when finished using find and replace.
+With this method only one line needs to be changed, syntax highlighting is usually preserved, and jumping to definition still works. It also has the advantage that it is extremely easy to remove all instances of `sorry;` from a file when finished using find and replace, and to quickly toggle this on and off to check things still work when desired.
 
 ### Memory and deterministic timeout
 Lean has two configurable flags for configuring its resource usage, memory and execution time.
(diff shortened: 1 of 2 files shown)


templates/tips_and_tricks.md
```diff
@@ -0,0 +1,49 @@
...
+Lean has two configurable flags for configuring its resource usage, memory and execution time.
+While the defaults for these are usually fairly sensible, when writing proofs it can be helpful to increase these options, if deterministic timeouts or out of memory issues slow down development.
+This is especially helpful when using tactics like `library_search` that even if they take a long time will be replaced by a faster tactic if they succeed.
+
+To change these settings we can add the flags `-M  <some number of megabytes>` (default 4096) for the memory limit and `-T <number of allocations>` (default 100000).
+In VSCode this can be configured by opening the settings and searching for `Lean Memory Limit` and `Lean Time Limit`.
+
+
+### Old mode
+If you change a file early on in the import hierarchy in a large project is changed and then attempt to work on a file that depends on it much later, then Lean will try to check all files in between to ensure they didn't break.
```
Julian, line 27: The "in a large project is changed" I think is an accidental insertion.

Julian, line 27:
```suggestion
If you change a file early on in the import hierarchy and then attempt to work on a file that depends on it much later, then Lean will try to check all files in between to ensure they didn't break.
```

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,49 @@
...
+This behaviour can cause the Lean server to become unresponsive, or simply slow down development when.
```
Julian, line 28:
```suggestion
This behaviour can cause the Lean server to become unresponsive, or simply slow down development.
```

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,49 @@
...
+You can also call lean with the "--old" flag to help in these situations, though it does come with some caveats.
```
Julian, line 29: Was there some first suggestion that was to go before this? Otherwise I think:

```suggestion
You can call lean with the "--old" flag to help in these situations, though it does come with some caveats.
```

Julian (commented):
These are great!

templates/tips_and_tricks.md
alexjbest: Nope just a copy-paste error, thanks for the fixes!

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,50 @@
...
+  ...
+  HUGE PROOF,
+  ... }
+```
+With this method only one line needs to be changed, syntax highlighting is usually preserved, and jumping to definition still works. It also has the advantage that it is extremely easy to remove all instances of `sorry;` from a file when finished using find and replace, and to quickly toggle this on and off to check things still work when desired.
+
+### Memory and deterministic timeout
+Lean has two configurable flags for configuring its resource usage, memory and execution time.
+While the defaults for these are usually fairly sensible, when writing proofs it can be helpful to increase these options, if deterministic timeouts or out of memory issues slow down development.
+This is especially helpful when using tactics like `library_search` that even if they take a long time will be replaced by a faster tactic if they succeed.
```
urkud, line 20: `library_search` should be a link to the tactic page.

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,50 @@
...
+
+To change these settings we can add the flags `-M  <some number of megabytes>` (default 4096) for the memory limit and `-T <number of allocations>` (default 100000).
+In VSCode this can be configured by opening the settings and searching for `Lean Memory Limit` and `Lean Time Limit`.
+
+
+### Old mode
+If you change a file early on in the import hierarchy and then attempt to work on a file that depends on it much later, then Lean will try to check all files in between to ensure they didn't break.
+This behaviour can cause the Lean server to become unresponsive, or simply slow down development.
+You can call lean with the "--old" flag to help in these situations, though it does come with some caveats.
+In VSCode you can enable it by opening the VSCode settings and searching for "Lean: extra options" then add the flag `--old`.
```
urkud, line 30: In Emacs you can customize the variable `lean-extra-arguments`.

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,50 @@
...
+The old flag should therefore not be used when consistency of the whole project is important (e.g. in CI), though it can speed up development.
+
+### The `#exit` command and `.`
+
+The command `#exit` will cause lean to stop reading the file at the location it is inserted.
+This can be helpful when working on proofs in the middle of a file with many results or slow results below the current location.
+This is related to switching to `check visible lines` mode, but gives the user more precise control over which parts of the file are slow.
+
+Likewise when working at the start of a file Lean will recheck the import lines, this can be prevented by placing a `.` character (which delimits lean declarations) on a newline after the import lines.
+This can also be useful in other situations when Lean wants to recheck things above the lemma or definition you are changing unnecessarily.
```
urkud, line 43: I usually add an (almost) empty module level docstring. I'll need to write it later anyway.

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,97 @@
+# Lean Tips and Tricks
+
+This document contains some tips for intermediate and advanced users of Lean.
+It is currently focussed around Lean 3.
+
+### Commenting out large proofs using `sorry`
+
+Sometimes commenting out chunks of a proof is helpful, often a convenient way to systematically do disable some proofs steps (rather than commenting out subsections) is to temporarily skip a long subproof (of a subgoal which has been surrounded by braces), you can do this by prepending `sorry;`
```
Julian, line 8:
```suggestion
Sometimes skipping subportions of a proof is helpful (e.g. to disable a portion which is slow while working on the rest). The straightforward way to do so is to comment out the lines which constitute the subportion, but if a subgoal has been surrounded by braces, you can also temporarily skip it by prepending `sorry;`
```

Julian (commented):
The first entry still sounds a bit awkward to me. Maybe this works?

templates/tips_and_tricks.md
alexjbest: Yeah I like this better too :)

templates/tips_and_tricks.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+### Old mode
+If you change a file early on in the import hierarchy and then attempt to work on a file that depends on it much later, then Lean will try to check all files in between to ensure they didn't break.
+This behaviour can cause the Lean server to become unresponsive, or simply slow down development.
+You can call lean with the "--old" flag to help in these situations, though it does come with some caveats.
+In VSCode you can enable it by opening the VSCode settings and searching for "Lean: extra options" then add the flag `--old`.
+For Emacs the variable `lean-extra-arguments` can be modified to add this flag.
+This causes lean not to recompile oleans for unedited files and just assume everything worked, so in this situation it is helpful to avoid recompilation.
+However you do then run the risk that your change broke or changed the meaning of some file in-between. Thus it most applicable when you are adding a new lemma/definition early in the hierarchy that you are sure won't break later files, rather than when you are changing existing definitions in an incompatible way.
```
adomani, line 34:
```suggestion
However you do then run the risk that your change broke or changed the meaning of some file in-between. Thus this suggestion mostly applies when you are adding a new lemma/definition early in the hierarchy that you are sure won't break later files, rather than when you are changing existing definitions in an incompatible way.
```

riccardobrasca (approved):

arthurpaulino (approved):

PR #245, opened 2022-01-24
doc(templates/extras): `calc` in tactic mode

Vierkantor:
Explain that `calc` also exists in tactic mode, functioning like `refine calc`.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/interactive.20transitivty.20proof/near/269144562)

### DIFF
--- a/templates/extras/calc.md
+++ b/templates/extras/calc.md
@@ -12,6 +12,17 @@ calc a = b + 1 : H1
 ...    = c + 1 : by rw H2
 ```
 
+`calc` is also available in tactic mode. You can leave `_`s to create a new goal:
+```lean
+example (a b c : ℕ) (H1 : a = b + 1) (H2 : b = c) : a = c + 1 :=
+begin
+  calc a = b + 1 : H1
+  ...    = c + 1 : _,
+  { rw H2 }
+end
+```
+In fact, `calc A = B : H ...` in tactic mode functions exactly like a call to `refine (calc A = B : H ...)`.
+
 ## Error messages, and how to avoid them
 
 Note that the error messages can be quite obscure when things aren't quite right, and often the red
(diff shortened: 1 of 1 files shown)


templates/extras/calc.md
```diff
@@ -12,6 +12,17 @@ calc a = b + 1 : H1
...
+`calc` is also available in tactic mode. You can leave `_`s to create a new goal:
+```lean
+example (a b c : ℕ) (H1 : a = b + 1) (H2 : b = c) : a = c + 1 :=
+begin
+  calc a = b + 1 : H1
+  ...    = c + 1 : _,
+  { rw H2 }
+end
+```
+In fact, `calc A = B : H ...` in tactic mode functions exactly like a call to `refine (calc A = B : H ...)`.
```
kmill, line 24: These ellipses are a little confusing since they're also part of the syntax for `calc`, but I'm not sure of a good solution... maybe this?
```suggestion
In fact, in tactic mode, `calc [...]` functions exactly like a call to `refine (calc [...])`.
```
Should we mention that you can do `exact calc [...]` if you want to override the `refine` behavior? (I'm not sure when you'd want to though!)

YaelDillies: You can use an ellipsis symbol!
```suggestion
In fact, `calc A = B : H …` in tactic mode functions exactly like a call to `refine (calc A = B : H …)`.
```

Vierkantor: I think the ellipsis symbol is still too likely to be confusing. As a third option, perhaps:
```suggestion
In fact, `calc [etc]` in tactic mode functions exactly like a call to `refine (calc [etc])`.
```

YaelDillies: The square brackets look like they are part of the syntax.

jcommelin: I think it all doesn't matter too much. Users likely already understand that part of the syntax. The context is making clear that what's important in this sentence is the beginning of the code example.

robertylewis:
I agree with @jcommelin that the discussion about ellipses doesn't matter too much. Readers will learn by example, and I think that sentence makes clear that it's an explanation, not a concrete example to copy.

PR #247, opened 2022-01-25
doc(data/maintainers): add Riccardo Brasca as maintainer

riccardobrasca:
Add myself as a new maintainer.


hrmacbeth:
I think you can say "Maître de conférences" if you like, no need to be entirely anglocentric :)

PR #249, opened 2022-01-30
Add glossary entries for "HoTT" and "kernel"

Julian:


templates/glossary.md
```diff
@@ -227,6 +242,22 @@ One of two closely related concepts:
 * A [declaration](#declaration) created with the eponymous `instance` command, or equivalently one marked with the `instance` [attribute](#attributes), either of which register the declaration with the typeclass inference system for use in the above.
 As a concrete example, [mathlib](#mathlib) defines an instance of `linear_order` for `ℝ`, enabling reals to be compared with `<`.
 
+### kernel
+
+In the context of Lean's implementation (and proof assistants more generally), the kernel is the central component which verifies the correctness of each proof.
+
+The goal of a proof assistant's kernel is to support all of the reasoning of the language with a small and bug-free, at the expense of expressivity or user-friendliness.
+Specifically, the kernel uses a simpler but less featureful representation of proofs -- one a Lean user would normally not interact with, but which is intended to prevent the possibility that a bug allows invalid proofs to be accepted.
+The user-facing portions of the Lean language are instead provided by its higher-level [modes](#mode), which ultimately are compiled into this internal kernel representation by the elaborator.
```
robertylewis, line 251: I don't love this description. It's not inaccurate to say that "input-Lean" is a different language from "kernel-Lean." "Less featureful representation of proofs" is more of a stretch. 

I guess what's bothering me is that it presents input- and kernel-Lean as side by side? Kernel-Lean is 100% the "real" language, there's nothing very principled about the input layer. Input-friendly proof sketches go into the elaborator which tries to turn them into real proofs which can be checked by the kernel. Users do interact with fully elaborated proofs before those proofs reach the kernel, too; it might be more accurate to decouple the kernel from fully elaborated CIC proofs.

templates/glossary.md
```diff
@@ -202,6 +202,21 @@ A heavy `rfl` occurs when `rfl` is asked to perform many steps of definitional r
...
+
+Lean 3's [kernel](kernel) introduced singleton elimination, which is inconsistent with HoTT's univalence axiom.
+As a result, core Lean 3 no longer shipped a HoTT library.
+A partial port of the Lean 2 HoTT library for Lean 3 was made in [an external project by Gabriel Ebner and fellow contributors](https://github.com/gebner/hott3).
+
+As an illustrative concrete example, there are many [definitions of the type of natural numbers](https://en.wikipedia.org/wiki/Natural_number#Formal_definitions) which can be viewed as constructing equivalent mathematical objects.
+Core Lean has a [Peano-esque implementation](https://leanprover-community.github.io/mathlib_docs/init/core.html#nat) of them, and [mathlib](#mathlib) has an additional [binary representation-based one](https://leanprover-community.github.io/mathlib_docs/data/num/basic.html#pos_num) which is shown to be equivalent.
+Given that Lean does *not* have the univalence axiom, these two types are equivalent, but they are not *equal* as types.
```
robertylewis, line 217:
```suggestion
Given that Lean does *not* have the univalence axiom, these two types are equivalent, but they are not provably *equal* as types.
```
It's consistent in Lean 3, but not provable, that any two types of the same cardinality are equal.

robertylewis (commented):
The HoTT description looks good!

templates/glossary.md
Julian: Thanks for the feedback. I suppose I meant "expressive" rather than "featureful"-- I was trying to paraphrase Andrej's post:

> The foundational system implemented in the kernel is too rudimentary to be directly usable for sophisticated mathematics. Instead, the user writes their input in a more expressive formal language 𝑉 (the vernacular) that is designed to be practical and useful. Typically 𝑉 is quite complex so that it can accommodate various notational conventions and other accepted forms of mathematical expression. A second component of the proof assistant, the elaborator, translates 𝑉 to 𝐹 and passes the translations to the kernel for verification.

But clearly I missed the nuances (and yeah featureful was a bad choice of words).

I'll try to give it another shot.

Julian: @robertylewis I've neutered and reworked the kernel entry a bit. Can you let me know if you have any further suggestions on it?

robertylewis:
This looks like a good summary for the level its targeting! Thanks again, and sorry for sleeping on this for so long.

PR #254, opened 2022-02-10
add events page

robertylewis:
@urkud suggested this on Zulip, so I figured I'd save him some work while I have a spare minute. There's likely more to fill in here. Feel free to suggest or push directly.


robertylewis:
image

PR #256, opened 2022-03-03
Add opensearch support to the mathlib docs

eric-wieser:
This should make supported browsers offer a "search mathlib docs" feature in the title bar.

It's not clear to me whether this will work in the mathlib doc repo, there are rumors online that it has to be at the site root.

I'll make a PR there anyway to test (https://github.com/leanprover-community/doc-gen/pull/161)

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -7,6 +7,7 @@
 
   <link rel="stylesheet" href="{{base_url}}/css/lean.css" >
   <link rel="shortcut icon" href="{{base_url}}/img/favicon.ico">
+  <link rel="search" type="application/opensearchdescription+xml" title="mathlib docs" href="/opensearch.xml">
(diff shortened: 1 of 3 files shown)


eric-wieser:
I can't really come up with a way of testing this other than deploying it and seeing what happens.

eric-wieser:
Thanks for merging this. Now that I can test it, it seems that this doesn't actually work in chrome :(

Maybe it works in firefox like @gebner said https://github.com/leanprover-community/doc-gen/pull/161 did? Either way, if it doesn't work in chrome we may as well revert this and put the change in doc-gen instead.

gebner:
I don't think this is deployed yet.  At least I can't find any mention of `opensearch.xml` in the page source code.

bryangingechen:
The jobs I re-ran just finished, so hopefully it's deployed now?

gebner:
Can confirm that it works on firefox.

PR #261, opened 2022-04-09
upgrade moment.min.js

bryangingechen:
Dependabot tells us there's a security alert in this version: https://github.com/advisories/GHSA-8hfj-j24r-96c4


gebner:
I don't think we were affected by the security issue. If I understand it correctly, you could only trigger it when loading locale files from the local filesystem (and we're using it in the browser).

AFAICT, we're using moment to format the timestamps in the statistics page.

PR #269, opened 2022-06-23
Update macos_details.md

unhyperbolic:
Looking at https://formulae.brew.sh/cask/elan and https://formulae.brew.sh/formula/elan-init, it seems that the command needs to change.


bryangingechen (approved):
Nice catch, thank you!

PR #271, opened 2022-07-12
Add row numbering to table mathlib_stats table

BoltonBailey:
I think this code will [number the rows in the table](http://tabulator.info/docs/5.3/format#formatter-rownum), so that we stop needing a column for "# by commits".


robertylewis:
#deploy

Ruben-VandeVelde:
I think https://github.com/leanprover-community/mathlib_stats/pull/10 will work better

BoltonBailey:
Ah, I agree completely. Thanks for figuring out the right place for this @Ruben-VandeVelde .

PR #276, opened 2022-08-01
Add some more naming conventions based on established practice

eric-wieser:


robertylewis (approved):

Vierkantor (approved):
All of the new symbols are indeed already established, so I don't expect any objection to adding them to the tables.

PR #277, opened 2022-08-03
add BIRS workshop

mattrobball:
Adding the BIRS workshop on formalization for cohomology theories. Future website 
is https://www.birs.ca/events/2023/5-day-workshops/23w5124.

### DIFF
--- a/data/events.yaml
+++ b/data/events.yaml
@@ -5,6 +5,13 @@
   end_date: August 16 2024
   type: conference
 
+- title: Formalization of cohomology theories
+  url: https://www.birs.ca/events/2023/5-day-workshops/23w5124
+  start_date: May 21 2023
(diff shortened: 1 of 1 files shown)


data/events.yaml
```diff
@@ -33,6 +33,13 @@
   location: Providence, RI, USA
   type: tutorial
 
+- title: Formalization of cohomology theories
+  url: https://www.birs.ca/events/2023/5-day-workshops/23w5124
+  start_date: May 21 2022
```
adamtopaz, line 38: The year should be 2023 (and presumably this should be somewhere higher up in the file to keep the chronological order)

mattrobball: Should be fixed. Ordered the events by `end_date`. `start_date` would provide a different ordering.

PR #278, opened 2022-08-11
add a notice about pip path

kokushkin:
Got an error "bash: pip3: command not found"
It wasn't clear where to find it)


semorrison:
As far as I can see, this error should not happen to users with "standard" systems who have followed the installation instructions, so I'm worried that it is more noise that helpful. I'd probably want to see evidence that more than one user is seeing this problem, and that this is the correct fix, before adding something like this.

PR #283, opened 2022-09-06
add a git config step to the mac os x instructions

collares:
I am not a Mac user, but [two users](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Installation.20on.20macOS) have reported Git errors while installing Homebrew, and [StackOverflow says](https://stackoverflow.com/questions/39836190/homebrew-install-failed-during-git-fetch-origin-masterrefs-remotes-origin-mas) this is due to the lack of a configured email address.


semorrison:
I am a little skeptical that the linked StackOverflow error is really diagnosing this correctly. Surely install homebrew without having configured an email address in `git` is extremely common??

Given we're about to rewrite these instructions, and hopefully actually demo them on a fresh (virtual) machine, I think I'd prefer not to merge this possibly spurious additional step.

PR #297, opened 2022-11-18
Update events.yaml

dagurtomas:
Add Copenhagen masterclass in June 2023

### DIFF
--- a/data/events.yaml
+++ b/data/events.yaml
@@ -1,3 +1,10 @@
+- title: Formalisation of mathematics
+  url: https://www.math.ku.dk/english/calendar/formalisation-of-mathematics/
+  location: University of Copenhagen, Denmark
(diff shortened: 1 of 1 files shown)


data/events.yaml
```diff
@@ -1,3 +1,10 @@
+- title: Formalisation of mathematics
+  url: https://www.math.ku.dk/english/calendar/formalisation-of-mathematics/
+  location: University of Copenhagen, Denmark
+  start_date: June 26 2023
+  end_date: June 30 2023
+  type: masterclass
```
PatrickMassot, line 6: This type of event won't be recognized. You need to choose between `conference`, `tutorial` and `workshop` as explained on top of https://leanprover-community.github.io/events.html.

dagurtomas: Changed to workshop

PR #300, opened 2022-12-12
Add LftCM24

riccardobrasca:


riccardobrasca:
We don't have a website yet... I put the link to the CIRM webpage otherwise building fails.

PR #302, opened 2022-12-21
Lean 3 and mathlib3

madvorak:


PatrickMassot:
I don't think we should change `mathlib` to `mathlib3` since this name isn't used anywhere else.

madvorak:
> I don't think we should change `mathlib` to `mathlib3` since this name isn't used anywhere else.

ah sry

templates/install/windows.md
```diff
@@ -1,6 +1,6 @@
 # Installing Lean and mathlib on Windows
```
PatrickMassot, line 1: Could you also write Lean 3 here?

templates/install/windows.md
```diff
@@ -80,7 +80,7 @@ This document describes using VS Code (for emacs, look at https://github.com/lea
    (or ![(image of icon)](img/extensions-icon.png) in older versions) in the side bar on the left edge of
    the screen (or press <kbd>Shift</kbd><kbd>Ctrl</kbd><kbd>X</kbd>) and search for `leanprover`.
 4. Select the `lean` extension (unique name `jroesch.lean`). There is also a
-   `lean4` extension, but that one does not work with mathlib.
+   `lean4` extension, but that one would require us to work with mathlib4.
```
PatrickMassot, line 83: That sentence is a bit weird. Maybe 
```suggestion
   `lean4` extension, but that one isn't relevant here.
```

PR #304, opened 2022-12-22
Add workshop on formalising algebraic geometry

jcommelin:


kbuzzard:
My only issue with this PR was that AIM made it clear that if we wanted to make the topic more broad, they would be happy with the idea, so we could still go for something like "formalising geometry". Still, I guess that if we decide to change the title we just make another PR :-)

PR #310, opened 2023-02-04
Updated Setup Instructions for Apple Silicon

MahdiKhaleghi:
We must distinguish between two brew setups to use and maintain them properly.


semorrison:
This is now obsolete. Everything just works, no need for special casing apple silicon.

PR #316, opened 2023-03-27
Added instructions with installing Python

qrico64:
I encountered the problem with "/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python", and turns out the problem was not selecting "Disable path length limit". I feel like adding this would help some people!


PatrickMassot:
I'm closing this since Python no longer have any relevance since we switched to Lean 4.

PR #317, opened 2023-05-02
Mention Lean 4

PatrickMassot:


templates/get_started.md
```diff
@@ -1,5 +1,9 @@
 # Get started with Lean
 
+This webpage is about the current stable version of Lean, which is Lean 3, although 
+the community is currently switching to Lean 4. 
+See the [Lean 4 manual](https://leanprover.github.io/lean4/doc/) for information about installing Lean 4.
```
eric-wieser, line 5: On the assumption this works
```suggestion
<div class="alert alert-info">

This webpage is about the current stable version of Lean, which is Lean 3, although 
the community is currently switching to Lean 4. 
See the [Lean 4 manual](https://leanprover.github.io/lean4/doc/) for information about installing Lean 4.

</div>
```
The website certainly uses bootstrap, I don't know for sure whether we parse HTML in our markdown.

PR #338, opened 2023-07-21
Link to mathlib4 decls in undergrad/overview/100 theorems

eric-wieser:
Depends on:
* https://github.com/leanprover-community/mathlib4/pull/6043


PatrickMassot:
@eric-wieser in case you didn't notice, this PR doesn't build. But thanks anyway!

eric-wieser:
I wrote a message to that effect on Zulip: we also need to export the line number information.

eric-wieser:
@PatrickMassot, this is now green. Happy to test in prod? Unfortunately after repeated crashes I'm hitting github rate limit errors, so can't test locally any further today.

robertylewis (approved):
LGTM!

PR #340, opened 2023-08-03
Add ICMS workshop

mariainesdff:
Add an entry for the "Formalisation of Mathematics:
Workshop for Women and Mathematicians of Minority Gender" strategic workshop at ICMS. The workshop website will be updated when available.

### DIFF
--- a/data/events.yaml
+++ b/data/events.yaml
@@ -75,6 +75,13 @@
   end_date: March 29 2024
   type: tutorial
 
+- title: "Formalisation of Mathematics: Workshop for Women and Mathematicians of Minority Gender"
+  url: https://www.icms.org.uk/
+  location: International Centre for Mathematical Sciences (ICMS), Edinburgh, United Kingdom
+  start_date: May 27 2024
+  end_date: May 31 2024
+  type: tutorial
+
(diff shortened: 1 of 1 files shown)


data/events.yaml
```diff
@@ -75,6 +75,13 @@
   end_date: March 29 2024
   type: tutorial
 
+- title: "Formalisation of Mathematics: Workshop for Women and Mathematicians of Minority Gender"
+  url: https://www.icms.org.uk/
+  location: International Centre for Mathematical Sciences (ICMS), Edinburgh, United Kingdom
+  start_date: May 27 2024
+  end_date: May 31 2024
+  type: tutorial
```
robertylewis, line 83: Just to double check: we use the "tutorial" tag for LftCM-style events where novices learn formalization, and "workshop" as a catch-all for most other things. (Maybe we should revisit the tag choices at some point.) Is "tutorial" accurate here?

mariainesdff: Yes, this is mainly an LftCM-style introductory workshop (although we will also welcome more experienced participants), so I think that the tutorial tag is accurate.

PR #341, opened 2023-08-03
add cs formalization conferences

robertylewis:
One reason we created the events page back in the day was to highlight CS "publication conferences" that weren't apparent to everyone. We've mised this summer cycle but this adds the past ones and CPP 2024, and let's add ITP and CICM 2024 when they're announced.

### DIFF
--- a/data/events.yaml
+++ b/data/events.yaml
@@ -1,3 +1,31 @@
+- title: Interactive Theorem Proving 2023
+  url: https://mizar.uwb.edu.pl/ITP2023/
+  start_date: July 31 2023
+  end_date: August 4 2023
+  location: Bialystok, Poland
+  type: conference
+
+- title: Conference on Intelligent Computer Mathematics 2023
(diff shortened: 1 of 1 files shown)


data/events.yaml
```diff
@@ -1,3 +1,31 @@
...
+  end_date: January 17 2023
+  location: Boston, MA, USA
+  type: conference
+
+- title: Certified Programs and Proofs 2024
+  url: https://popl24.sigplan.org/home/CPP-2024
+  start_date: January 15 2024
+  end_date: January 16 2024
+  location: London, UK
+  type: conference
```
eric-wieser, line 27: Are these intended to be in chronological order?

eric-wieser (approved):
Feel free to merge with or without re-ordering.

data/events.yaml
robertylewis: They get sorted when the page is generated, and the yaml list is out of order anyway, so I didn't bother!

PR #343, opened 2023-08-08
add new maintainers, photos, bios to the maintainers page

j-loreaux:
--- 

I have only added myself here for now, but I thought other new maintainers would just pile onto this PR with their info.


data/teams.yaml
```diff
@@ -91,6 +94,7 @@
     - Kexing Ying
     - Aaron Anderson
     - Kevin Buzzard
+    - Eric Rodriguez
```
eric-wieser, line 97: @ericrbg; do you want a middle name here?

ericrbg: My second surname Boidi would be great :)

eric-wieser: Feel free to add a suggestion in github, then I'll commit and merge

ericrbg, line 97:
```suggestion
    - Eric Rodriguez Boidi
```

PR #344, opened 2023-08-09
update `naming.md` to reflect Lean 4 style

j-loreaux:


templates/contribute/naming.md
```diff
@@ -70,76 +111,23 @@ When translating the statements of theorems into words, this dictionary is often
...
-- `le_of_mul_le_mul_right`
-
-An injectivity lemma that uses "left" or "right" should refer to the
-argument that "changes". For example, a lemma with the statement
-`a - b = a - c ↔ b = c` could be called `sub_right_inj`.
-
-We can also use the word "self" to indicate a repeated argument:
-
-- `mul_inv_self`
-- `neg_add_self`
```
eric-wieser, line 142: Don't these largely still apply?

templates/contribute/naming.md
```diff
@@ -401,7 +389,7 @@ and it is reasonable to update and replace these as you come across them.
 Note however that constructors for inductive types have
 automatically generated unidirectional implications, named `.inj`,
 and there is no intention to change this.
-When such an automatically generated lemma already exists,
+xWhen such an automatically generated lemma already exists,
```
eric-wieser, line 392:
```suggestion
When such an automatically generated lemma already exists,
```

templates/contribute/naming.md
j-loreaux: I removed them because I think they are largely copied below already (let me know if you think otherwise)

PR #348, opened 2023-08-17
feat (Lean courses around the world): add M&M

madvorak:


data/courses.yaml
```diff
@@ -230,3 +230,18 @@
...
     As projects in 2022 students created formalizations of exercise solutions for my book "Automatic complexity",
     some of which are included at https://github.com/bjoernkjoshanssen/diophantine-lemma
+- name: Programování a dokazování v Leanu
+  instructor: Martin Dvorak
+  location: online
+  website: https://mam.mff.cuni.cz/
+  repo: https://github.com/madvorak/lean-mam
+  tags: ['functional programming', 'intro to proofs']
+  summary: >
+    We are a (korespondenční seminář)[https://cs.wikipedia.org/wiki/Koresponden%C4%8Dn%C3%AD_semin%C3%A1%C5%99] (lit. "correspondence seminar" in the sense of
```
eric-wieser, line 240: This markdown isn't quite right:
```suggestion
    We are a [korespondenční seminář](https://cs.wikipedia.org/wiki/Koresponden%C4%8Dn%C3%AD_semin%C3%A1%C5%99) (lit. "correspondence seminar" in the sense of
```

madvorak: Thanks for noticing!

data/courses.yaml
```diff
@@ -303,6 +303,21 @@
...
     some of which are included at https://github.com/bjoernkjoshanssen/diophantine-lemma
+- name: Programování a dokazování v Leanu
+  instructor: Martin Dvorak
+  location: online
+  website: https://mam.mff.cuni.cz/
+  repo: https://github.com/madvorak/lean-mam
+  tags: ['functional programming', 'intro to proofs']
+  summary: >
+    We are a [korespondenční seminář](https://cs.wikipedia.org/wiki/Koresponden%C4%8Dn%C3%AD_semin%C3%A1%C5%99) (lit. "correspondence seminar" in the sense of
+    a seminar arranged by exchanging letters and emails) called (M\&M)[https://cs.wikipedia.org/wiki/M%26M_(%C4%8Dasopis)], which is something like
```
eric-wieser, line 314: This link is backwards too

madvorak: Is it now OK with the two `)` in a row?

madvorak: Oh, a backslash is needed!

PR #349, opened 2023-08-21
Fix the doc links on {100,undergrad,overview} pages

eric-wieser:
For now this doesn't attempt to fix the broken links *within* the 100.yaml problems.


make_site.py
```diff
@@ -223,7 +225,8 @@ def replace_link(name, id):
         return '/mathlib4_docs/' + name
     else:
         try:
-            return declarations[name].docLink
+            # note: the `.bmp` data files use doc-relative links
+            return  '/mathlib4_docs/' + declarations[name].docLink
```
eric-wieser, line 229:
```suggestion
            return '/mathlib4_docs/' + declarations[name].docLink
```

make_site.py
```diff
@@ -208,8 +208,10 @@ class DeclarationDataEntry:
                     continue
                 doc_decls.append(DocDecl(
                     name=decl,
+                    # TODO: add missing `/mathlib4_docs/` prefix to links within this header
                     decl_header_html = header_data.get(decl, ''),
-                    docs_link=decl_info.docLink,
+                    # note: the `.bmp` data files use doc-relative links
+                    docs_link= '/mathlib4_docs/' + decl_info.docLink,
```
eric-wieser, line 214:
```suggestion
                    docs_link='/mathlib4_docs/' + decl_info.docLink,
```

PR #352, opened 2023-08-23
links changed

madvorak:


PatrickMassot:
Thanks Martin. I am a bit worried that this would create a web page that is updated to Lean 4 but crucially linked to the style page that is still completely Lean 3. Do you think you could work on this one? 

I'm also pinging @j-loreaux who updated the naming page.

madvorak:
Is the main problem the following?

Adhere to the guidelines:
The [style guide](https://github.com/leanprover-community/leanprover-community.github.io/blob/f7aa0ee0b484b64bd7c4d66fd935e6a0ff9238e9/templates/contribute/style.html) for contributors.
The explanation of [naming conventions](https://github.com/leanprover-community/leanprover-community.github.io/blob/f7aa0ee0b484b64bd7c4d66fd935e6a0ff9238e9/templates/contribute/naming.html).
The [documentation guidelines](https://github.com/leanprover-community/leanprover-community.github.io/blob/f7aa0ee0b484b64bd7c4d66fd935e6a0ff9238e9/templates/contribute/doc.html).

j-loreaux:
Yeah, I think it doesn't make sense to update a page on the website in a piecemeal fashion. Any page that gets updated should be subject to a complete overhaul (even if much of the information ends up being similar). For instance, this page still makes mention of `leanproject` in several places that are not changed, and I'm not sure we even want to suggest forking anymore since we never use it due to our CI setup. There's probably a lot that needs to be revisited.

As for the style guide, I actually have that mostly rewritten already. I'll try to get a PR up tonight or tomorrow for that.

madvorak:
I don't want to work on a complete overhaul. Should I close this PR?
Somebody else will hopefully do it in an organized manner.

PatrickMassot:
Ok,  let's close it for now.

PR #355, opened 2023-08-29
Corrected a link

matematiflo:
The only relevant part of this PR is the file templates/install/macos.md, where I corrected the link, which does not work because of the markdown syntax.

The issue is that the link in question does not work when one clicks on it on the following webpage:

https://leanprover-community.github.io/lean3/install/macos.html


PatrickMassot:
It seems this PR has a serious git issue and is probably not what you intended.

PR #357, opened 2023-08-30
update style guide to Lean 4

j-loreaux:


templates/contribute/style.md
```diff
@@ -116,157 +98,171 @@ See our [documentation requirements](doc.html) for more suggestions and examples
...
-After stating the theorem, we generally do not indent the first line
-of a proof, so that the proof is "flush left" in the file.
+After stating the theorem, we indent the lines in the subesequent proof by 2 spaces.
 ```lean
-open nat
-theorem nat_case {P : nat → Prop} (n : nat) (H1 : P 0) (H2 : ∀ m, P (succ m)) : P n :=
-nat.induction_on n H1 (assume m IH, H2 m)
+open Nat
+theorem nat_case {P : Nat → Prop} (n : Nat) (H1 : P 0) (H2 : ∀ m, P (succ m)) : P n :=
+  Nat.recOn n H1 (fun m IH ↦ H2 m)
```
PatrickMassot, line 115: Isn't this only for tactic proofs?

templates/contribute/style.md
```diff
@@ -116,157 +98,171 @@ See our [documentation requirements](doc.html) for more suggestions and examples
...
-      absurd H6 H2)))),
-[...]
+example (n k : Nat) (h : n < k) : ... :=
+  have h1 : n ≠ k := by exact ne_of_lt h
+  ...
+
+example (n k : Nat) (h : n < k) : ... :=
+  have h1 : n ≠ k := by
+    exact ne_of_lt h
+  ...
```
PatrickMassot, line 208: It would be nice to have an example which doesn't encourage unnecessary `by exact`.

templates/contribute/style.md
```diff
@@ -282,184 +278,221 @@ example (n : ℝ) (h : 1 < n) : 0 < n := by linarith
...
 ```lean
-example (n : ℝ) : 1 < n → 0 < n := λ h, by linarith
+example (n : ℝ) : 1 < n → 0 < n := fun h ↦ by linarith
 ```
 
 and
 
 ```lean
-example (n : ℕ) : 0 ≤ n := dec_trivial
+example (n : ℕ) : 0 ≤ n := Nat.zero_le n
```
PatrickMassot, line 287: Isn't there a Lean 4 analogue of `dec_trivial`? I know `decide` is slightly different.

templates/contribute/style.md
```diff
@@ -282,184 +278,221 @@ example (n : ℝ) (h : 1 < n) : 0 < n := by linarith
...
 ```
 
+### The anonymous function arrow
+
+Lean has several nice syntax options for declaring anonymous functions. For very simple
+functions, one can use the centered dot as the function argument, as in `(· ^ 2)` to 
+represent the squaring function. However, sometimes it is necessary to refer to the 
+arguments by name (e.g., if they appear in multiple places in the function body). The
+Lean default for this is `fun x => x * x`, but mathlib introduces an alternative 
+notation which we prefer to use in mathlib: `fun x ↦ x * x` (inserted with `\mapsto`).
```
PatrickMassot, line 311: This is not introduced in mathlib, it comes from core Lean. The only thing that mathlib does is to use the option to print this notation.

templates/contribute/style.md
j-loreaux: I think `decide` *is* supposed to be the Lean 4 analogue of `dec_trivial`, but it doesn't seem to handle free or meta variables (at least, that's the error message I get when I try `by decide`).

templates/contribute/style.md
j-loreaux: Ah thanks, I had that wrong. I think what I said is correct now?

templates/contribute/style.md
j-loreaux: Isn't what only for tactic proofs? This web page? No, the old version had plenty of term-mode, and most of the term-mode that exists here I adapted directly from that.

PR #359, opened 2023-08-31
update the contribution page to Lean 4 and add commit conventions

j-loreaux:
This does two things:

1. adds a page addressing the commit conventions (i.e., PR title and description conventions) for mathlib, as the one we used to link to was in the Lean (not Lean 4!) repository. That document was copied, and then edited it slightly to reflect mathlib better, including the optional `(<scope>)` listing the files changed.
2. updates the contribution page to Lean 4, including a link to the new commit conventions page.

### DIFF
--- a/templates/contribute/commit.md
+++ b/templates/contribute/commit.md
@@ -1,0 +1,95 @@
+# Pull request title and description conventions
+
+We are using the following convention for writing pull request titles and descriptions.
+
+## Format
+
+Note: "Title:" and "Description:" do not actually appear
+
+```markdown
+  Title:
+  <type>(<optional-scope>): <subject>
+
+  Description:
+  <body>
+  <NEWLINE>
+  <footer>
+  <NEWLINE>
+  <dependencies>
+```
+
+`<type>` is:
+
+ - feat (feature)
+ - fix (bug fix)
+ - doc (documentation)
+ - style (formatting, missing semicolons, ...)
+ - refactor
+ - test (when adding missing tests)
+ - chore (maintain)
+ - perf (performance improvement, optimization, ...)
+
+`<optional-scope>` is a name of module or a directory which contains changed modules.
+This is not necessary to include, but may be useful if the `<subject>` is insufficient.
+The `Mathlib` directory prefix is always omitted.
+For instance, it could be
+
+- Data/Nat/Basic
+- Algebra/Group/Defs
+- Topology/Constructions
+
+`<subject>` has the following constraints:
+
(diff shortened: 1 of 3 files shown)


templates/contribute/index.md
```diff
@@ -97,8 +72,7 @@ It's also okay to let our central CI servers do this for you.
   otherwise a yellow circle if CI is still working, or a red cross if something went wrong.
   Click on the red cross to see details.)
   You can also check CI status on the command line by installing `hub` and running `hub ci-status`.
```
PatrickMassot, line 74:
```suggestion
  You can also check CI status on the command line by installing [`hub`](https://hub.github.com/) and running `hub ci-status`.
```

PatrickMassot (commented):
Thanks!

templates/contribute/commit.md
```diff
@@ -0,0 +1,95 @@
...
+
+- just as in ``<subject>``, use imperative, present tense
+- include motivation for the change and contrast with previous
+  behavior
+
+`<footer>` is optional and may contain two items:
+
+- Breaking changes: All breaking changes have to be mentioned in
+  footer with the description of the change, justification and
+  migration notes
```
eric-wieser, line 57: Do we actually care about this? I don't think I've seen this used in a commit message for mathlib before

templates/contribute/commit.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+an example with dependent PRs:
+
+```markdown
+feat: The norm on `Unitization` is a C⋆-norm
+
+This shows that C⋆-algebras are always `RegularNormedAlgebra`s, so that their `Unitization` is equipped with a norm. Moreover, we show this norm is a C⋆-norm.
+
+- [ ] depends on: #5330
```
eric-wieser, line 91: Right now our PR template asks for these too go _below_ the `---` (ie not in the commit message at all).

Are you advocating a change to put these in a different order?

templates/contribute/commit.md
j-loreaux: I think we actually do, although we're bad about enforcing it. There was talk in some Zulip thread which I can't find at the moment about doing exactly this so we can easily find it later. I think there was even discussion about the exact format it should take, but I don't think it was finalized.

templates/contribute/commit.md
j-loreaux: oof, you're right, but I've *always* put them before the `---` break, and it seems most other people do too. I guess I'm not sure which we should advocate for, but you're correct that as currently listed this is a departure from the PR description template.

PR #360, opened 2023-09-05
feat(project.md): add windows troubleshooting

fpvandoorn:
also fix 2 tiny typos.

### DIFF
--- a/templates/extras/conv.md
+++ b/templates/extras/conv.md
@@ -140,7 +140,7 @@ conversion mode using the `for` command. The following performs rewriting
 only for the second and third occurrences of `b * c`:
 
 ```lean
-example (a b c : ℕ) : (b * c) * (b * c) * (b * c) = (b * c) * (c * b)  * (c * b):=
+example (a b c : ℕ) : (b * c) * (b * c) * (b * c) = (b * c) * (c * b) * (c * b):=
 by conv { for (b * c) [2, 3] { rw mul_comm } }
 ```
 
--- a/templates/install/project.md
+++ b/templates/install/project.md
@@ -86,7 +86,7 @@ The keyword `math` at the end of this command adds `mathlib4` to the dependencie
   `New file`, and type a filename to create a file there.
 
 If you want to make sure everything is working, you can start by
-creating, say `my_project/MyProject/test.lean` containing:
+creating, say `my_project/MyProject/Test.lean` containing:
 ```lean
 import Mathlib.Topology.Basic
 
--- a/templates/install/project.md
+++ b/templates/install/project.md
@@ -67,6 +67,12 @@ The keyword `math` at the end of this command adds `mathlib4` to the dependencie
 
 * Go inside the `my_project` folder and type `lake update`, then `lake exe cache get` and then `mkdir MyProject`.
 
+  * Some Windows users have reported an error like this:
+    ```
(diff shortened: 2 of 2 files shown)


templates/install/project.md
```diff
@@ -67,6 +67,12 @@ The keyword `math` at the end of this command adds `mathlib4` to the dependencie
 
 * Go inside the `my_project` folder and type `lake update`, then `lake exe cache get` and then `mkdir MyProject`.
 
+  * Some Windows users have reported an error like this when running `lake exe cache get`:
+    ```
+      curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - The revocation function was unable to check revocation for the certificate
+    ```
+    If you see this error, you likely have an antivirus program that is unhappy that `lake` downloads the compiled files from mathlib. You can temporarily disable your antivirus program, or tell it to ignore/allow the operation of `lean.exe`. After you do this, run `lake exe cache get!`. The exclamation mark forces `lake` to re-download the cache files it failed to download before.
```
eric-wieser, line 74: I think this is confusing two separate issues:
* Antivirus is messing with all of Lean's IO, slowing it to a crawl
* Antivirus is messing with curl's certificate checking, causing it to complain that the connection has been compromised

The solution is for the first, but the error message is from the second.

fpvandoorn: Better now?

templates/install/project.md
```diff
@@ -67,6 +67,14 @@ The keyword `math` at the end of this command adds `mathlib4` to the dependencie
 
 * Go inside the `my_project` folder and type `lake update`, then `lake exe cache get` and then `mkdir MyProject`.
 
+  * Some Windows users have reported an error like this when running `lake exe cache get`:
+    ```
+      curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - The revocation function was unable to check revocation for the certificate
+    ```
+    If you see this error, you likely have an antivirus program that scans each downloaded file, which results in errors. Please disable your antivirus program and then run
+    `lake exe cache get!`. The exclamation mark forces `lake` to re-download the cache files it failed to download before.before running this command. (If you are uncomfortable disabling your antivirus, try to follow [these instructions](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/lake.20exe.20cache.20get.20errors/near/389019448) and then run `lake exe cache get!`).
```
eric-wieser, line 75: Double `before`
```suggestion
    `lake exe cache get!`. The exclamation mark forces `lake` to re-download the cache files it failed to download before running this command. (If you are uncomfortable disabling your antivirus, try to follow [these instructions](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/lake.20exe.20cache.20get.20errors/near/389019448) and then run `lake exe cache get!`).
```

templates/extras/conv.md
```diff
@@ -140,7 +140,7 @@ conversion mode using the `for` command. The following performs rewriting
 only for the second and third occurrences of `b * c`:
 
 ```lean
-example (a b c : ℕ) : (b * c) * (b * c) * (b * c) = (b * c) * (c * b)  * (c * b):=
+example (a b c : ℕ) : (b * c) * (b * c) * (b * c) = (b * c) * (c * b) * (c * b):=
```
eric-wieser, line 143:
```suggestion
example (a b c : ℕ) : (b * c) * (b * c) * (b * c) = (b * c) * (c * b) * (c * b) :=
```

PR #361, opened 2023-09-06
Documentation: additional necessary command on Apple Silicon Install

jlefkoff:
When installing for an M2 Mac, homebrew will not automatically install as x86_64 unless you specify it using the `arch -x86_64` command. This is outlined [here](https://gist.github.com/progrium/b286cd8c82ce0825b2eb3b0b3a0720a0). Adding this to the front of the homebrew install will make it install at `/usr/local/bin` as intended.


semorrison:
@jlefkoff, I think on the contrary no special treatment is needed at all, and you can just follow the standard instructions whether you are on Apple Silicon or Intel.

I'll close this, but please ping me on leanprover.zulipchat.com, or here, if I am wrong.

PR #365, opened 2023-09-20
Remove the lake +nightly... instruction in project.md

semorrison:
Remove the lake +nightly... instruction, as it is no longer needed.


semorrison:
Could someone verify this independently?

fpvandoorn:
I had a student last week that had Lean 3 installed. `lake` was not a recognized command on the command line until you are in a repository that specifies a Lean 4 version. So I think this sentence is still helpful for some people.

You could still keep your change and add "if that doesn't work, you probably have an old version of `lake`. In that case, use ..."

semorrison:
No, let's keep it for a while then.

PR #367, opened 2023-09-23
Add Rome event details to events.yaml

oliver-butterley:


oliver-butterley:
I notice that the scheduled build workflow failed [https://github.com/leanprover-community/leanprover-community.github.io/actions/runs/6287865482/job/17072755243]

I suspect this is my fault because the name of our event includes a semicolon and I didn't quote it in the yaml.

PR #369, opened 2023-09-25
feat(templates/install): warning for distro package managers

lambda-fairy:
It's implied by the rest of the docs, but I think it's worth making it explicit, given how often the problem appears on Zulip.

### DIFF
--- a/templates/install/debian_details.md
+++ b/templates/install/debian_details.md
@@ -43,7 +43,7 @@ reading the bash script that will be downloaded below:
 
 It might be tempting to use your Linux distribution's package manager (APT, DNF, AUR, etc.) to install Lean instead. **Don't do this!**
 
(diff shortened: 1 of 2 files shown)


kbuzzard:
Maybe you should add that at the time of writing lean is making new releases every few weeks, so it's highly likely that the version in your package manager is already out of date?

lambda-fairy:
Thanks @kbuzzard, added.

Also added a note about elan handling multiple versions. (I hedged my point with "most" to placate the Nix fans)

PatrickMassot:
I don't understand the placement of those paragraph, and I don't understand the last sentence. 

And I'm skeptical about the general idea. People who arrive at those web page already decided they were not using their package managers.

jcommelin:
I agree with Patrick that I don't understand the last sentence.

Otoh, I don't think it hurts to add this warning: I often skim install pages in combination with installation via distro package managers.

PatrickMassot:
I am closing this since it did not converge.

PR #370, opened 2023-09-25
indentation for `|`

mo271:


mo271:
https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/style.20for.20proofs.20with.20.60.7C.60.2E

PR #373, opened 2023-09-26
make photo 100x100

mattrobball:


mattrobball:
Apologies for another clean up. I think this is the last one.

PR #377, opened 2023-10-11
Make the lean3 warning more obvious

eric-wieser:


templates/_base.html
```diff
@@ -47,10 +47,10 @@
 <div id="mainbox">
 
   <main>
-    <div class="alert alert-info">
+    <div class="alert alert-warning">
```
eric-wieser, line 50: Or maybe even
```suggestion
    <div class="alert alert-danger">
```

PR #400, opened 2023-11-22
Make sure that lean.bib is always normalized

sorawee:
This PR adds a GitHub workflow that checks that `lean.bib` is normalized. 

The PR also brings the normalization instruction to the top. Unfortunately, `bibtool` will associate the instruction as a comment for the first bib entry, so a newer entry that precedes `AGLST23` (e.g., `AAA`) will still displace the instruction, and will require a manual fix again. However, this hopefully will be rare.

### DIFF
--- a/.github/workflows/bibtool.yml
+++ b/.github/workflows/bibtool.yml
@@ -1,0 +1,33 @@
+name: check bib
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Thanks!

PR #402, opened 2023-11-23
space after left arrow

mo271:


templates/contribute/style.md
```diff
@@ -503,6 +503,9 @@ example {x y : ℝ} (hxy : x ≤ y) (h : ∀ ε > 0, y - ε ≤ x) : x = y :=
     linarith  
 ```
 
+Both when using the rewrite tactic (`rw`) and the `do` notation, there should
+be a space after the left arrow (`←`).
```
eric-wieser, line 507: Can you give an example of each?

mo271: Let's not mention the do notation, since this moslty relevant in tactics.
I added an example, not sure if it is a good one...

templates/contribute/style.md
```diff
@@ -503,8 +503,16 @@ example {x y : ℝ} (hxy : x ≤ y) (h : ∀ ε > 0, y - ε ≤ x) : x = y :=
     linarith  
 ```
 
-Both when using the rewrite tactic (`rw`) and the `do` notation, there should
-be a space after the left arrow (`←`).
+When using the rewrite tactic `rw`, there should be a space after the left arrow (`←`):
+
+```lean
```
eric-wieser, line 508: I was thinking of a much shorter inline example like `rw [← add_comm a b]`

eric-wieser: And for the do notation you could use `do (← f) + (← g)`

mo271: Done

semorrison (commented):
Please mention simp along with rw.

Also, the second sentence missing a verb.

mo271:
> Please mention simp along with rw.
> 
> Also, the second sentence missing a verb.

Done!

PR #429, opened 2024-01-05
use zulip data in map

robertylewis:


make_site.py
```diff
@@ -455,12 +457,48 @@ class User:
...
+               if not user['is_bot'] and user['is_active'] and validate_coordinates(user)] if 'ZULIP_KEY' in os.environ else []
+
+users = [
+    User(
+        fullname=user['full_name'], 
+        lon=float(get_user_field(user, 'longitude')), 
+        lat=float(get_user_field(user, 'latitude')), 
+        github=get_user_field(user, 'github'),
+        website=get_user_field(user, 'website')) 
+    for user in zulip_users]
```
eric-wieser, line 501: Perhaps:
```suggestion
def get_users():
  if client is None:
    return

  for user in client.get_members({"include_custom_profile_fields": True})['members']:
    if user['is_bot'] or not user['is_active']:
      continue
    try:
      lat = float(get_user_field(user, 'latitude'))
      lon = float(get_user_field(user, 'longitude'))
    except Exception:
      continue
    yield User(
        fullname=user['full_name'], 
        lon=float(get_user_field(user, 'longitude')), 
        lat=float(get_user_field(user, 'latitude')), 
        github=get_user_field(user, 'github'),
        website=get_user_field(user, 'website'))
  
users = list(get_users())
```
to avoid getting the coordinates twice.

eric-wieser: Because I don't know how many times we iterate over it below, and I figured this was  least likely to break anything.
Indeed, if we're careful we could leave it as an iterator.

make_site.py
```diff
@@ -455,12 +457,48 @@ class User:
...
+    'latitude': '5148',
+    'longitude': '5149',
+    'github': '3658'
+}
+
+def get_user_field(user, field):
+    try:
+        return user['profile_data'][profile_data_fields[field]]['value']
+    except KeyError:
+        return None
```
eric-wieser, line 480: nit:
```suggestion
    field_id = profile_data_fields[field]
    try:
        return user['profile_data'][field_id]['value']
    except KeyError:
        return None
```
so that typos in field names aren't silenced.

eric-wieser (approved):
Feel free to merge to see if this works.

PR #437, opened 2024-02-03
add two references and fix one

faenuccio:
Fix one ref to an ITP paper and add two more


robertylewis:
I'm confused about why CI succeeded on this PR -- the build on the main branch is failing because of dulicated references. @faenuccio once the site has built and deployed, could you check that all the references you expect are being displayed at https://leanprover-community.github.io/papers.html ?

faenuccio:
Sure, I will keep on eye and will report as soon as I see that the site has built.

faenuccio:
OK, I have double-checked and there is no duplicate. There was one problem with a curly bracket (fixed in #438 ) but I think that the duplicate warning comes from the fact that there are two submissions with almost the same name (but for capitalization) that are actually two different papers.

PR #439, opened 2024-02-05
Modification to "Book" paragraph in Learning Lean 4 page

motib:
Add link to my tutorial "A Gentel Introduction to Lean"
Rewrote the paragraph "Books" for consitency and clarity, also changing the name to "Documents".

### DIFF
--- a/templates/learn.md
+++ b/templates/learn.md
@@ -20,26 +20,32 @@ There is no point learning Lean 3 at this stage.
   using Lean, and then independent topic files about elementary analysis,
   abstract topology and mathematical logic.
 
-## Books
+## Documents
 
-* If you prefer reading a book (with exercises), the standard mathematics oriented reference is
-  [Mathematics in Lean](https://leanprover-community.github.io/mathematics_in_lean/).
-  You can also download it 
+* [Mathematics in Lean](https://leanprover-community.github.io/mathematics_in_lean/) by Jeremy Avigad and 
+  Patrick Massot is the standard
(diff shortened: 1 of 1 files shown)


turibe:
I believe that "How to prove it with Lean" still uses Lean 3?

PatrickMassot:
@turibe, thanks for your help, but actually that project was ported to Lean 4, see https://github.com/djvelleman/HTPILeanPackage/tree/master.

However this project is pretty different from the ones that are already mentioned and the maintainer team already decided not to include it in this specific list. We will think about the other document and report back.

turibe:
Ah, I was judging by https://djvelleman.github.io/HTPIwL/IntroLean.html, which is perhaps older. (All such pages should ideally mention which version of Lean they refer to / were tested on, but many don't.)

turibe:
> Thanks, I was hastily judging by https://djvelleman.github.io/HTPIwL/IntroLean.html, which does not mention a version. (All such pages should ideally mention which version of Lean they refer to / were tested on, but many don't... it gets confusing.) p.s. Looking closer, I do see one version hint there : "“Lean 4: Infoview: Toggle” :-) , so that's Lean 4 too.

jcommelin:
@motib Yesterday we had a triage meeting for pull requests to the website. We decided that we want to use this list for resources that have an established track record in the community of being a good resource for beginners.

We are happy that you wrote your introduction to Lean, but we do not think that it should be included in this list.

PR #447, opened 2024-03-02
Add Michael Stoll to reviewers in teams.yaml, sort list

MichaelStollBayreuth:
Following a request by @kbuzzard, I have taken the opportunity to sort the list of reviewers alphabetically by last name.


MichaelStollBayreuth:
(The list on https://leanprover-community.github.io/teams/reviewers.html that is generated from this is sorted by *first* names, however...)

PatrickMassot:
Indeed the reason why we didn't care about the order in this file is that the list is ordered at use time.

PR #451, opened 2024-03-12
Link to Zulip profiles from the map

eric-wieser:
Not yet tested.


PatrickMassot:
Could you please test this change?

eric-wieser:
I always find it a disproportionate amount of effort to test website changes vs just merging and fixing issues... I can't remember what failed last time I tried a local build.

PR #454, opened 2024-03-22
Update project.md

riccardobrasca:
`lake build` seems to automatically do `lake exe cache get` now (at least on my computer).


fpvandoorn:
(you meant `lake update` instead of `lake build` in the PR description)

PR #467, opened 2024-04-24
feat: changes to the learn page

fpvandoorn:
* Add `Formalising Mathematics` book
* Add `The Hitchhiker's Guide to Logical Verification` book (this should not go under `(Meta)-programming and tactic writing` since it's still about proving, not programming).
* Change description of `Theorem Proving in Lean`
* Remove meetings that were not aimed at beginners, but link to the event page.

### DIFF
--- a/templates/learn.md
+++ b/templates/learn.md
@@ -4,13 +4,13 @@ There are many ways to start learning Lean, depending on your background and
 taste. They are all fun and rewarding, but also difficult and
 occasionally frustrating. Proof assistants are still difficult to use,
 and you cannot expect to become proficient after one afternoon of
-learning. Note that all resources listed on that page are about Lean 4. 
+learning. Note that all resources listed on that page are about Lean 4.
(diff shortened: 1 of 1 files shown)


fpvandoorn:
Should I also add [How to Prove it With Lean](https://djvelleman.github.io/HTPIwL/How-To-Prove-It-With-Lean.pdf)?

kbuzzard:
I guess the CI failure is https://github.com/microsoft/linux-package-repositories/issues/130 which should be fixed now. The changes look fine to me. Why are you deleting some but not all of the pre-2022 conferences?

fpvandoorn:
I kept the meetings that were aimed at beginners, and removed the ones that weren't. We can remove more if you think that is appropriate.

kbuzzard:
I made a suggestion about changing the description of my notes to course notes rather than a book, but other than that this looks good to me. Are you happy with the change? If so we can merge this.

kbuzzard:
Thanks!

PR #472, opened 2024-05-01
Create speedup.md

MichaelStollBayreuth:
A first version of a document explaining how one can (try to) make a Mathlib file build faster. See [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Ways.20to.20speed.20up.20Mathlib/near/435991508).

### DIFF
--- a/templates/contribute/speedup.md
+++ b/templates/contribute/speedup.md
@@ -13,7 +13,7 @@ the experiment carried out in Mathlib PR
    instructions below.
 3. The previous step can potentially be repeated with a lowered setting of
    `profiler.threshold`, to find and speed up things that are not very slow,
-   but also not very fast. This will eventually deliver diminshing returns, however.
+   but also not very fast. This will eventually deliver diminishing returns, however.
 4. Remove the profiler options again and PR!
 
 ## Dealing with specific slow steps
--- a/templates/contribute/speedup.md
+++ b/templates/contribute/speedup.md
@@ -1,0 +1,88 @@
+# How to speed up a Mathlib file
+
+We explain how a slow Mathlib file can be made faster. These notes are based on
+the experiment carried out in Mathlib PR
+[#12412](https://github.com/leanprover-community/mathlib4/pull/12412).
+
+1. The first step is to find out which parts of the code are slow. To do so, 
+   add a line `set_option profiler true` after the `import` statements.
+   This will produce lines of the form `<blah> took <number>ms` (or even `<number>s`)
+   in the infoview, recording steps that took at least `100ms` (the lower bound in ms
+   can be adapted via `set_option profiler.threshold <num>`).
+2. For each such line, try to make the corresponding step faster following the
+   instructions below.
+3. The previous step can potentially be repeated with a lowered setting of
+   `profiler.threshold`, to find and speed up things that are not very slow,
+   but also not very fast. This will eventually deliver diminshing returns, however.
+4. Remove the profiler options again and PR!
+
+## Dealing with specific slow steps
+
+Here we explain how one can try to speed up various parts of the code that cause
+the profiler to produce messages in the infoview.
+
+### `typeclass inference of <name> took <a long time>`
+
+1. Add `set_option trace.Meta.synthInstance true in` immediately before the declaration
+   causing the message.
+2. Look at the generated instance synthesis trace in the infoview and find the instance(s)
+   of `<name>` that are slow.
+3. Use `#synth <name> <args>` before the declaration (possibly temporarily adding `variable`s
+   to the context if needed) to obtain a suitable term providing the instance.
+4. Add a line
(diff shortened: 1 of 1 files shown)


templates/contribute/speedup.md
```diff
@@ -0,0 +1,88 @@
...
+1. The first step is to find out which parts of the code are slow. To do so, 
+   add a line `set_option profiler true` after the `import` statements.
+   This will produce lines of the form `<blah> took <number>ms` (or even `<number>s`)
+   in the infoview, recording steps that took at least `100ms` (the lower bound in ms
+   can be adapted via `set_option profiler.threshold <num>`).
+2. For each such line, try to make the corresponding step faster following the
+   instructions below.
+3. The previous step can potentially be repeated with a lowered setting of
+   `profiler.threshold`, to find and speed up things that are not very slow,
+   but also not very fast. This will eventually deliver diminshing returns, however.
```
adomani, line 16:
```suggestion
   but also not very fast. This will eventually deliver diminishing returns, however.
```

templates/contribute/speedup.md
```diff
@@ -0,0 +1,88 @@
...
+Replace the relevant `simp/simpa` call by `simp?/simpa?` and click on the `Try this:`
+suggestion to replace it by a `simp/simpa only` call. In some cases it is also possible
+to prune the list of lemmas to some extent.
+
+**Trade-off:** Proofs can get several dense lines longer.
+
+### `elaboration took <a long time>`
+
+Look for `_`s in the declaration that triggers it, find out what they are filled by,
+and replace them by the corresponding explicit arguments.
```
adomani, line 61: `show_term` can help with this (or can be too verbose!).

adomani (approved):
This looks great, thanks!

templates/contribute/speedup.md
MichaelStollBayreuth: I was thinking of `_`s in the statement (in this case one can see what they are filled with by putting the cursor after the `by`). But I assume this can also be an issue within proofs. Does `show_term` work with terms as opposed to tactics? Or would you have to write `show_term exact <term>`?

adomani: I wouldn't be surprised is a term-mode `show_term` existed, but it do not know what it is anyway.

templates/contribute/speedup.md
```diff
@@ -0,0 +1,88 @@
+# How to speed up a Mathlib file
+
+We explain how a slow Mathlib file can be made faster. These notes are based on
+the experiment carried out in Mathlib PR
+[#12412](https://github.com/leanprover-community/mathlib4/pull/12412).
+
+1. The first step is to find out which parts of the code are slow. To do so, 
+   add a line `set_option profiler true` after the `import` statements.
```
eric-wieser, line 8: I'm surprised not to see `trace.profiler` mentioned here; do you have a goof feeling for the difference?

templates/contribute/speedup.md
```diff
@@ -0,0 +1,88 @@
...
+
+Try to replace the slow tactic by calls to simpler ones.
+
+* For example, a slow `nontriviality ... using ...` can be replaced by
+```lean
+  rcases subsingleton_or_nontrivial ... with H | H
+  · -- get `Subsingleton` case out of the way
+    ...
+  -- now we have `Nontrivial ...`
+```
```
eric-wieser, line 83:
````suggestion
  ```lean
  rcases subsingleton_or_nontrivial ... with H | H
  · -- get `Subsingleton` case out of the way
    ...
  -- now we have `Nontrivial ...`
  ```
````

File: templates/contribute/speedup.md [hunk 5]
```diff
@@ -0,0 +1,88 @@
+# How to speed up a Mathlib file
+
+We explain how a slow Mathlib file can be made faster. These notes are based on
+the experiment carried out in Mathlib PR
+[#12412](https://github.com/leanprover-community/mathlib4/pull/12412).
+
+1. The first step is to find out which parts of the code are slow. To do so, 
+   add a line `set_option profiler true` after the `import` statements.
+   This will produce lines of the form `<blah> took <number>ms` (or even `<number>s`)
```
grunweg, line 9: I just tried your guide when creating #13454. First of all, thanks for writing it - really helpful!

When I first used this, the output seemed not as helpful as it could be: if I get something like
simp took 2.01s
tactic execution of Lean.Parser.Tactic.rintro took 103ms
simp took 905ms
simp took 507ms
simp took 1.26s
simp took 1.14s
elaboration took 645ms

is there an easy way to correlate the output with the input lines? Sure, I can manually count the simp invocations, perhaps fiddle with the thresholds so all `simp`s are shown etc. - but is there a nicer way?

grunweg:
I have another comment, which is more of a wish "can you also answer this": do you know how I can address output "linting to <a long time>"?

MichaelStollBayreuth:
No idea. I don't even know what the linters do that run by default...

PR #473, opened 2024-05-04
Fix: complete the removal of 3 pages which were removed in prior commit 

oliver-butterley:
In commit [892d908](https://github.com/leanprover-community/leanprover-community.github.io/commit/892d9086699a5676dbc4d24d7909c0ac9b1be212) three files were removed. However the deploy system doesn't remove these files from the deployed site so they are still visible. They were no longer visible in the contents but one could arrive there by searching. 

This PR manually removes these three files and completes the work of the mentioned commit.

(An alternative would be for the workflow to be updated to deal with deleted files automatically but this wasn't immediately clear to me.)


PatrickMassot:
Thanks!

PR #478, opened 2024-05-31
Add bit about upgrade project.md

rwst:
Please add/change if inaccurate.

### DIFF
--- a/templates/install/project.md
+++ b/templates/install/project.md
@@ -117,6 +117,13 @@ It's not necessary for using Lean.
 
 See [How to contribute to mathlib](https://leanprover-community.github.io/contribute/index.html).
 
+## Updating the toolchain in your project
+
+Eventually you might want to follow Lean development and upgrade the toolchain in your project.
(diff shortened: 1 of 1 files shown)


fpvandoorn:
If the project depends on Mathlib *and* the old version of lake is able to parse the new lakefiles of all dependencies, then this happens automatically, IIUC. Maybe this should be mentioned, because in most cases nothing has to be done by the user.

rwst:
@fpvandoorn You mean it's not necessary to `lake update`, just change `lean-toolchain`?

fpvandoorn:
I only was talking about manually updating `lean-toolchain`.

fpvandoorn:
I don't think these instructions are very good, users don't have to directly interact with `elan` themselves. I added my own version here: #487 487

rwst:
No problem. I don't have much experience with lean. I'm glad that you are taking over, and that there will be good instructions on that page. Closing.

fpvandoorn:
Thanks for taking the initiative in writing these instructions. They were indeed needed!

PR #487, opened 2024-06-07
add updating and troubleshooting instructions, remove hosting

### LINKED ISSUE #489: Broken link in `/install/project.html`
There is a broken link in [this ](https://leanprover-community.github.io/install/project.html#hosting-your-project-on-github) section, it gives a 404 error: https://leanprover-community.github.io/ci.html.

fpvandoorn:
* Add detailed update instructions
* The hosting section was mostly there to link to `ci.html` which was removed.
* Add troubleshooting instructions when a corrupted Lean version was downloaded
* Add instructions to clean-up old Lean versions/Mathlib cache.

fixes #489

### DIFF
--- a/templates/install/project.md
+++ b/templates/install/project.md
@@ -66,7 +66,7 @@ commands should be typed in a terminal.
 The keyword `math` at the end of this command adds `mathlib4` to the dependencies of your project, so that you can use `import Mathlib` in your project files.
 
 * Go inside the `my_project` folder and type `lake update` and then `mkdir MyProject`.
-  * Windows users seeing a `curl: (35) schannel: next InitializeSecurityContext failed` error should read [this note](#initializesecuritycontext-error-on-windows).
+  * Windows users seeing a `curl: (35) schannel: next InitializeSecurityContext failed` error should read [this note](#troubleshooting).
 
 * Launch VS Code, either through your application menu or by typing
   the `code .` command.
@@ -104,22 +104,28 @@ Also, you can get the Lean documentation inside VS Code using
 inside the text field that appears, type "lean doc" and hit <kbd>Enter</kbd>.
 Then click "Mathematics in Lean" or "Theorem Proving in Lean" and enjoy.
 
-## Hosting your project on GitHub
+## Updating Mathlib in your project
 
-Your project is already a git repository, and as it grows,
-you may want to host it on [GitHub](https://guides.github.com/activities/hello-world/).
-If you take this step, the community offers some
-[GitHub Actions scripts](../ci.html) that could help manage your repository.
-But don't worry if you don't know what this means.
-It's not necessary for using Lean.
+If you have a project depending on Mathlib, and you want to update to the latest version of Lean and Mathlib, you have to do two steps:
+* Update the `lean-toolchain` file in your repository to the latest version. You can do this automatically by running the following in the root directory of your repository.
+```
+curl https://raw.githubusercontent.com/leanprover-community/mathlib4/master/lean-toolchain > lean-toolchain
+```
+* run `lake update`. This will update Lean, Mathlib and download the new Mathlib cache for you.
+
+You should then check whether all files in your repository still compile. The most convenient way to do this is to have a root file that imports all files in your repository. You can do this by running the following (replace `MyProject` twice by your project name):
+```
+find MyProject -name "*.lean" | env LC_ALL=C sort | sed 's/\.lean//;s,/,.,g;s/^/import /' > MyProject.lean
+```
+After that, you can run `lake build` to build all files in your repository.
 
 ## Contributing to mathlib
 
 See [How to contribute to mathlib](https://leanprover-community.github.io/contribute/index.html).
 
-## InitializeSecurityContext error on Windows
+## Troubleshooting
 
-Some Windows users have reported an error like this when running `lake exe cache get`:
+* Some Windows users have reported an error like this when running `lake exe cache get`:
 
 ```
   curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - The revocation function was unable to check revocation for the certificate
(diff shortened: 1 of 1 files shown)


templates/install/project.md
```diff
@@ -104,22 +104,30 @@ Also, you can get the Lean documentation inside VS Code using
...
-If you take this step, the community offers some
-[GitHub Actions scripts](../ci.html) that could help manage your repository.
-But don't worry if you don't know what this means.
-It's not necessary for using Lean.
+If you have a project depending on Mathlib, and you want to update to the latest version of Lean and Mathlib, you have to do two steps:
+* Update the `lean-toolchain` file in your repository to the latest version. You can do this automatically by running the following in the root directory of your repository.
+```
+curl https://raw.githubusercontent.com/leanprover-community/mathlib4/master/lean-toolchain > lean-toolchain
+```
+* run `lake update`. This will update Lean, Mathlib and download the new Mathlib cache for you.
```
utensil, line 114: It's preferable to use `lake -R -Kenv=dev update`. `-R` avoids lake to use an old lakefile.olean. `-Kenv=dev` avoids removing doc-gen4 from the manifest if the user is using it.

See discussions on Zulip [1](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Recommend.20way.20to.20add.20mathlib.20master.20as.20lib.3F/near/438097875), [2](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/UnicodeBasic.20error.20when.20upgrading.20to.204.2E8.2E0rc2/near/441776050).

templates/install/project.md
```diff
@@ -104,22 +104,30 @@ Also, you can get the Lean documentation inside VS Code using
...
+
+You should then check whether all files in your repository still compile. The most convenient way to do this is to have a root file that imports all files in your repository. You can do this by running the following (replace `MyProject` twice by your project name):
+```
+find MyProject -name "*.lean" | env LC_ALL=C sort | sed 's/\.lean//;s,/,.,g;s/^/import /' > MyProject.lean
```
utensil, line 118: I thought now it's preferable to use `lake exe mk_all`? See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/automatically.20generate.20ProjectName.2Elean).

fpvandoorn: Thanks for the link. I have issues with `mk_all`, but if that works reliably, we can adapt to that.

templates/install/project.md
fpvandoorn: Thanks for the pointers. 
I missed the `-R` option in my own repositories using doc-gen, so thanks for that piece of information. These instructions are not really for using doc-gen, but I guess adding the general case doesn't hurt.

templates/install/project.md
fpvandoorn: It seems that there is an error with this script on Windows. I suggest we use the current instructions for now, and we can simplify them once that issue is fixed.

fpvandoorn: I mention `lake exe mk_all` now with the caveat that it doesn't work on Windows.

templates/install/project.md
```diff
@@ -104,22 +104,33 @@ Also, you can get the Lean documentation inside VS Code using
...
-If you take this step, the community offers some
-[GitHub Actions scripts](../ci.html) that could help manage your repository.
-But don't worry if you don't know what this means.
-It's not necessary for using Lean.
+If you have a project depending on Mathlib, and you want to update to the latest version of Lean and Mathlib, you have to do two steps:
+* Update the `lean-toolchain` file in your repository to the latest version. You can do this automatically by running the following in the root directory of your repository.
+```
+curl -L https://raw.githubusercontent.com/leanprover-community/mathlib4/master/lean-toolchain -o lean-toolchain
+```
+* run `lake -R -Kenv=dev update`. This will update Lean, Mathlib and download the new Mathlib cache for you.
```
semorrison, line 114: Why are you suggesting `-R -Kenv=dev`? This seems unnecessarily complicated. We shouldn't be putting end users in a situation where either of these are required, no?

utensil: Because without -R, the user could end up running an outdated lakefile.olean, without -Kenv=dev, the user could end up removing doc-gen4 accidentally.

These 2 paper cuts are still the most frequent ones for every Lean/Mathlib bump. And they still dodges issues in the era of 4.8.0-rc*.

Until a raw `lake update` stops creating look-for-help topics on Zulip, this is worth mentioned in the doc.

Recommend the users to run a raw `lake update` then guide them to troubleshooting to add these 2 options back will be wasting the time of users.

semorrison: Why is it a problem that they remove doc-gen4? Users should never see that! I am skeptical here.

semorrison: Okay, I have responded at that first thread asking people to not make life difficult. :-)

So far, I remain convinced that these instructions should not include workarounds for badly set up projects.

fpvandoorn: It's the de facto standard setup for projects downstream for Mathlib.
I'm happy to see this situation improved, and then simplifying these instructions, but I think the current instructions are useful in 10+ projects at the moment.

semorrison: Okay, let's proceed for now.

fpvandoorn: so shall we merge this?

templates/install/project.md
```diff
@@ -104,22 +104,33 @@ Also, you can get the Lean documentation inside VS Code using
...
+
+You should then check whether all files in your repository still compile.
+The most convenient way to do this is to run `lake exe mk_all`. However, that might not work on Windows at the moment.
```
grunweg, line 117: As I understand it, mk_all works on Windows now - so we can remove the `find` incantation, right?

fpvandoorn:
I now mention `lake exe mk_all` (without alternatives) and moved `lake -R -Kenv=dev update` into a parenthetical.

PR #492, opened 2024-06-13
doc: add hint about restarting computer.

joneugster:


joneugster:
This comes from a [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Install.20lean4.20on.20windows) where a Windows user got stuck trying to call `source ~/.bash_profile` on their Windows machine. Restarting is completely over-the-top, but might be simple enough and easy to understand.

templates/install/project.md
```diff
@@ -29,7 +29,7 @@ Open a terminal.
 
 * If you have not logged in since you installed Lean and mathlib, then
   you may need to first type `source ~/.profile` or
-  `source ~/.bash_profile` depending on your OS.
+  `source ~/.bash_profile` depending on your OS. (or simply restart your computer)
```
jcommelin, line 32:
```suggestion
  `source ~/.bash_profile` depending on your OS.
  (If you are on Windows, and don't know how to do this, another option is to restart your computer.)
```

PR #493, opened 2024-06-19
chore: update MacOS instructions to remove Homebrew references

mattrobball:
Homebrew `elan-init` lags in version and can be confusing to find. Change the detailed install instructions to reference the released installers.

### DIFF
--- a/templates/install/macos_details.md
+++ b/templates/install/macos_details.md
@@ -18,8 +18,10 @@ Installing `elan` and mathlib supporting tools
 1.  Grab the [latest release](https://github.com/leanprover/elan/releases/latest) for your architecture,
     unpack it, and run the contained installation program.
 
-    The installation will tell you where it will install elan to (~/.elan by default),
-    and also ask you about editing your shell config to extend PATH. elan can be uninstalled via elan self uninstall, which should revert these changes.
+    The installation will tell you where it will install `elan` to (`~/.elan` by default),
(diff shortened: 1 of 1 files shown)


templates/install/macos_details.md
```diff
@@ -15,17 +15,13 @@ all users.
...
 ---
 
-1.  Install [Homebrew](https://brew.sh/) if you do not already have it installed.
+1.  Grab the [latest release](https://github.com/leanprover/elan/releases/latest) for your architecture,
+    unpack it, and run the contained installation program.
 
-2.  Run `brew install elan-init` in a terminal window to
-    install `elan`.
+    The installation will tell you where it will install elan to (~/.elan by default),
+    and also ask you about editing your shell config to extend PATH. elan can be uninstalled via elan self uninstall, which should revert these changes.
```
semorrison, line 22:
```suggestion
    The installation will tell you where it will install `elan` to (`~/.elan` by default),
    and also ask you about editing your shell config to extend `PATH`. `elan` can be uninstalled via `elan self uninstall`, which should revert these changes.
    
(We discourage using the `homebrew` provided `elan-init` package, as users often find that this lags behind updates to the official release. We especially discourage using the `homebrew` formula named simply `lean`, which installs a fixed version of Lean.)
```

semorrison:
I think we should explicitly discourage using homebrew still, otherwise people will.

If you're happy with my suggested change (or not!) please go ahead and merge.

PR #496, opened 2024-07-04
enable rendering mermaid graphs

jcommelin:
- **wip**
- **Revert "wip"**
- **Enable rendering mermaid git graphs**

### DIFF
--- a/mistletoe_renderer.py
+++ b/mistletoe_renderer.py
@@ -104,12 +104,13 @@ class CustomHTMLRenderer(HTMLRenderer):
 
     def render_block_code(self, token):
         # replace math before highlighting
+        code = replace_math(token.children[0].content, self.math)
+        # render mermaid code separately so that the javascript renderer can detect it properly
         if token.language == 'mermaid':
             s = '<pre class="mermaid">'
-            s += token.children[0].content
+            s += code
             s += '</pre>'
             return s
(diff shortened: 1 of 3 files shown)


mistletoe_renderer.py
```diff
@@ -104,6 +104,11 @@ def wrap(self, source):
 
     def render_block_code(self, token):
         # replace math before highlighting
+        if token.language == 'mermaid':
+            s = '<pre class="mermaid">'
+            s += token.children[0].content
+            s += '</pre>'
+            return s
         code = replace_math(token.children[0].content, self.math)
```
bryangingechen, line 112: I think it's better to run this after the `replace_math` because the input to `render_block_code` has already had `remove_math` run on it (see line 57 in render_md), so there's a remote possibility we end up spitting out placeholder tokens instead of the original input string here. (In practice I don't think it will happen because LaTeX syntax and Mermaid syntax don't seem to overlap, but better safe than sorry.)
```suggestion
        code = replace_math(token.children[0].content, self.math)
        # render mermaid code separately so that the javascript renderer can detect it properly
        if token.language == 'mermaid':
            s = '<pre class="mermaid">'
            s += code
            s += '</pre>'
            return s
```

bryangingechen:
Thanks!

PR #504, opened 2024-07-20
Fix documentation for calc

riccardobrasca:
Removing the senteces "Kevin thinks".

The explanation on how `calc` works internally seems confusing and too complicated for this guide, so I am removing it.

### DIFF
--- a/templates/extras/calc.md
+++ b/templates/extras/calc.md
@@ -97,20 +97,8 @@ theorem T2 (a b c d : ℕ)
 What is actually going on here? The proofs themselves are not a mystery,
 for example `Nat.succ_le_succ h2` is a proof of `b + 1  ≤ c + 1`. The
 clever part is that lean can put all of these together to correctly
-deduce that if `U = V < W ≤ X < Y` then `U < Y`. The way this is done,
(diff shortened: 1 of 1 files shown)


templates/extras/calc.md
```diff
@@ -121,7 +109,7 @@ is `=`. Lean knows
 #check trans_rel_left -- {α : Sort u} {a b c : α} (r : α → α → Prop) (h₁ : r a b) (h₂ : b = c) : r a c
 ```
 
-and (Kevin believes) uses them if one of the operators is an equality operator. If however neither
+and uses them if one of the operators is an equality operator. If however neither
```
riccardobrasca, line 112: Note that I am not sure the sentence is correct, but anyway in practice it works like it were.

fpvandoorn: Yes, these are [instTransEq](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#instTransEq) and [instTransEq_1](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#instTransEq_1) (not sure if they have the highest priority, but there shouldn't be any conflicting instances).

fpvandoorn:
Seems good.

PR #508, opened 2024-07-28
make sure the example copyright header would pass CI

edegeltje:
the linter checking the header matches on `Authors:` and makes sure the line doesn't end in `.`, so it would fail on the example. also make the year current, while we're at it.


adomani:
Maybe you could also mention in the text below that you should use `Authors` even if there is only one and that there should be no `and` separating the penultimate and the last author.

edegeltje:
i took the freedom of suggesting to use commas rather than just discouraging the use of `and`.

adomani (approved):
Thanks for this!

I can see that there seem to be ending spaces to some of your new lines.  Apart from this, LGTM!

bryangingechen:
Thank you!

PR #510, opened 2024-08-02
doc: add tips about lake usage

joneugster:
Adding some notes about custom setups that are possible with a `lakefile.lean`, in particular how to use a single shared `mathlib` in multiple projects.

This might be useful for people with little disk space as well as for teaching setups where it might be reasonable to have a single `mathlib` on a shared drive and then have students reference this instead of downloading their own copy.

### DIFF
--- a/templates/install/tricks.md
+++ b/templates/install/tricks.md
@@ -3,7 +3,7 @@
 These tips and tricks about managing Lean projects should be considered workarounds.
 Some care is adviced when trying these non-standard setups.
 
-*Note:* Things here might change as `lake` is being developed, as features described here are not necessarily officially supported by `lake`. This file has been written for Lean `v4.10.0`. If in doubt, ask for help on [Zulip](https://leanprover.zulipchat.com).
+*Note:* Things here might change as `lake` is being developed, as features described here are not necessarily officially supported by `lake`. This file has been written for Lean `v4.16.0`. If in doubt, ask for help on [Zulip](https://leanprover.zulipchat.com).
 
 ## Shared Mathlib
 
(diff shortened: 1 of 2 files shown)


templates/install/tricks.md
```diff
@@ -0,0 +1,148 @@
...
+
+Here is a guide on best practise on how to achieve that.
+
+1) First, clone the version of mathlib somewhere on your computer:
+   ```bash
+   git clone --branch v4.10.0 git@github.com:leanprover-community/mathlib4.git
+   ```
+   Note that `v4.10.0` is the tag of the latest release, you can look at [mathlib's tags](https://github.com/leanprover-community/mathlib4/tags) to find out which is the most recent release version.
+
+   (If you don't have git setup correctly using an SSH key, you might want to use `git clone --branch v4.10.0 https://github.com/leanprover-community/mathlib4.git` instead.)
```
grunweg, line 21: Alternative phrasing, for your consideration:
```
The above line assumes you have set up git using an SSH key. If you have not set this up correctly, you may want to use `git clone --branch v4.10.0 https://github.com/leanprover-community/mathlib4.git` instead.```

File: templates/install/tricks.md [hunk 2]
```diff
@@ -0,0 +1,148 @@
...
+2) Go inside your mathlib and download cache:
```
grunweg, line 22: "download the current cache"?

templates/install/tricks.md
```diff
@@ -0,0 +1,148 @@
...
+   cd mathlib
+   lake exe cache get
+   ```
+3) If you ever want to **update** your global mathlib, come back to the mathlib directory and call
+   ```bash
+   git checkout v4.11.0
+   lake exe cache get
+   ```
+   with the [version](https://github.com/leanprover-community/mathlib4/tags) you'd like to update to.
+4) Now if you don't already have a Lean project, create it
```
grunweg, line 33:
```suggestion
4) If you don't already have a Lean project, create it.

File: templates/install/tricks.md [hunk 4]
```diff
@@ -0,0 +1,148 @@
...
+   * In `lakefile.lean`, replace the line `require "leanprover-community" / "mathlib"` with
+     ```
+     require mathlib from ".." / "relative" / "path" / "to" / "mathlib4"
+     ```
+   * Now inside `MyProject` you need to clean up lake:
+     ```bash
+     rm -rf .lake # because `lake clean` does not remove `.lake/packages/mathlib` which might have been downloaded by `lake new`.
+     lake clean # or potentially `lake update -R mathlib` instead
+     ````
+     *(note: it looks like a bug that with a simple `lake clean`, there might still be a folder `.lake/packages/mathlib` floating around from before you changed the `lakefile.lean`. However, deleting `.lake/` is a reasonably save action as it only contains build artifacts that are fully recovered by the next `lake` call.)*
```
grunweg, line 49:
```suggestion
     *(note: it looks like a bug that with a simple `lake clean`, there might still be a folder `.lake/packages/mathlib` floating around from before you changed the `lakefile.lean`. However, deleting `.lake/` is a reasonably safe action as it only contains build artifacts that are fully recovered by the next `lake` call.)*

File: templates/install/tricks.md [hunk 5]
```diff
@@ -0,0 +1,148 @@
...
+   <!-- 1) Make sure `lean-toolchain` contains the string `leanprover/lean4:v4.10.0` with the same version your shared mathlib is at. -->
+   * In `lakefile.lean`, replace the line `require "leanprover-community" / "mathlib"` with
+     ```
+     require mathlib from ".." / "relative" / "path" / "to" / "mathlib4"
+     ```
+   * Now inside `MyProject` you need to clean up lake:
+     ```bash
+     rm -rf .lake # because `lake clean` does not remove `.lake/packages/mathlib` which might have been downloaded by `lake new`.
+     lake clean # or potentially `lake update -R mathlib` instead
+     ````
```
grunweg, line 48:
```suggestion
     ```

File: templates/install/tricks.md [hunk 6]
```diff
@@ -0,0 +1,148 @@
...
+   lake exe cache get
+   ```
+   with the [version](https://github.com/leanprover-community/mathlib4/tags) you'd like to update to.
+4) Now if you don't already have a Lean project, create it
+   ```bash
+   lake new MyProject math.lean
+   cd MyProject
+   ```
+5) In the project `MyProject` you need to modify two things:
+   <!-- 1) Make sure `lean-toolchain` contains the string `leanprover/lean4:v4.10.0` with the same version your shared mathlib is at. -->
```
grunweg, line 39: Would you like to include this again?

templates/install/tricks.md
```diff
@@ -0,0 +1,148 @@
...
+     ```
+     require mathlib from ".." / "relative" / "path" / "to" / "mathlib4"
+     ```
+   * Now inside `MyProject` you need to clean up lake:
+     ```bash
+     rm -rf .lake # because `lake clean` does not remove `.lake/packages/mathlib` which might have been downloaded by `lake new`.
+     lake clean # or potentially `lake update -R mathlib` instead
+     ````
+     *(note: it looks like a bug that with a simple `lake clean`, there might still be a folder `.lake/packages/mathlib` floating around from before you changed the `lakefile.lean`. However, deleting `.lake/` is a reasonably save action as it only contains build artifacts that are fully recovered by the next `lake` call.)*
+   * Your project should be ready and when you add `import Mathlib` in a file and click "Restart File" in VSCode, it should be reasonably quick without rebuilding mathlib.
```
grunweg, line 50:
```suggestion
   * Your project should be ready: when you add `import Mathlib` in a file and click "Restart File" in VSCode, it should be reasonably quick, without rebuilding mathlib.

**Comment by joneugster** [comment]
Thank you @adomani for the review! I completely forgot this PR, but now I've addressed all your suggestions

File: templates/install/tricks.md [hunk 8]
```diff
@@ -0,0 +1,160 @@
...
+   * In `lakefile.lean`, replace the line `require "leanprover-community" / "mathlib"` with
+     ```
+     require mathlib from ".." / "relative" / "path" / "to" / "mathlib4"
+     ```
+   * Now inside `MyProject` you need to clean up lake:
+     ```bash
+     rm -rf .lake # because `lake clean` does not remove `.lake/packages/mathlib` which might have been downloaded by `lake new`.
+     lake clean # or potentially `lake update -R mathlib` instead
+     ```
+     *(note: it looks like a bug that with a simple `lake clean`, there might still be a folder `.lake/packages/mathlib` floating around from before you changed the `lakefile.lean`. However, deleting `.lake/` is a reasonably safe action as it only contains build artifacts that are fully recovered by the next `lake` call.)*
```
tydeu, line 61:
```suggestion
     ```bash
     lake update -R mathlib # use the new path for mathlib
     rm -rf .lake/packages/mathlib # delete the previous local clone of mathlib
     ```
```

`lake update -R mathlib` should be enough to switch to the relative mathlib. If not, that is a bug. Also, `lake clean` does not delete Lake packages, it merely removes build outputs.

templates/install/tricks.md
```diff
@@ -0,0 +1,160 @@
...
+   * Your project should be ready: when you add `import Mathlib` in a file and click "Restart File" in VSCode, it should be reasonably quick, without rebuilding mathlib.
+6) When you updated your global mathlib it might be enough to call
+   ```
+   lake update -R mathlib
+   ```
+   which would in theory update everything automatically.
+   However, if there are breaking changes to the `lakefile` parsing, you might need to
+   * edit `lean-toolchain` to be the same as your global mathlib.
+   * make sure `lakefile.lean` parses without error in the new version.
+   * try `lake update -R mathlib` again.
```
tydeu, line 71:
```suggestion
6) Updating your global mathlib should, in theory, update everything automatically.
   However, if there are breaking changes to the `lakefile` parsing, you might need to
   * edit `lean-toolchain` to be the same as your global mathlib.
   * make sure `lakefile.lean` parses without error in the new version.
   * try whatever broke again.
```

You should not need to `lake update` a relative dependency. (as path dependencies are not versioned).

tydeu (commented):
This PR was mentioned [again](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/How.20to.20Efficiently.20Use.20Mathlib.20in.20Multiple.20Projects.3F/near/486065997) on Zulip today, so I thought I should take another look. Below are some comments I had on on the shared mathlib section. In regards to the local versions section, the next version of Lake should, hopefully, have some built-in support local versions of dependency, making that section much less hacky. :-)

templates/install/tricks.md
joneugster: It is necessary, but for a different reason: mathlib has various post-update hooks which are called here. I added a sentense explaining this.

templates/install/tricks.md
joneugster: updated

joneugster:
@tydeu Would you mind having another look? Btw, if you do give me pointers to a better setup to switch between local and remote package versions, I'm happy to keep this guide updated.

We are using this setup in `lean4game` and I'm happy to implement any suggestions there and have them reflected here.

joneugster:
Or, maybe, I'll just close this PR and add these tricks to https://github.com/leanprover-community/mathlib-manual/ instead. Might be easier to keep them updated there...

joneugster:
I've now added these notes to https://leanprover-community.github.io/mathlib-manual/html-multi/.

PR #513, opened 2024-08-20
Prominently mention the API documentation 

omentic:
It took me surprisingly long to find https://leanprover-community.github.io/mathlib4_docs/, and I think it's one of the best references a newcomer to Lean can rely on. I wrote a tiny blurb (feel free to reword) in the hands-on section mentioning this: previously, the only link to the API documentation was a single small one on the sidebar.

### DIFF
--- a/templates/learn.md
+++ b/templates/learn.md
@@ -24,6 +24,12 @@ Some have Lean 3 versions, but there is no point learning Lean 3 at this stage.
   using Lean, and then independent topic files about elementary analysis,
   abstract topology and mathematical logic.
 
+* If you wish to learn directly from source, the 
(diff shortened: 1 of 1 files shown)


omentic:
Also - I'm aware that `leanprover-community` and `leanprover` are separate communities, but it would be nice if https://leanprover-community.github.io/mathlib4_docs/ could be mentioned on https://lean-lang.org/documentation/, since it not only hosts mathlib but also hosts core language documentation (Lean, Init, Std...)

fpvandoorn:
Looks good to me!

omentic:
Thanks!

PR #515, opened 2024-08-21
fix(people.yaml): Yury's employer

urkud:


data/people.yaml
```diff
@@ -141,7 +141,7 @@
 -
   name: Yury G. Kudryashov
   descr: |
-       Yury is a post-doc in mathematics at the Texas A&amp;M University.
+       Yury is a Formalization Lead at [Harmonic](https://harmonic.fun/about).
```
eric-wieser, line 144: I don't know for sure whether markdown works here. Perhaps merge this when you have some time to correct it if something goes wrong with the rendering?

urkud: I'm going to merge it, see if it works, then drop the link, if it doesn't.

PR #516, opened 2024-08-22
Add Loogle to sidebar links

omentic:
I stumbled across Loogle recently. It doesn't appear to be mentioned on the learning resources page yet.

This also changes the link to mathlib4_docs from "API documentation" to "General documentation". I think this is a lot clearer (and is what mathlib4_docs calls itself): when I was first poking around with Lean I thought this was some sort of FFI documentation, and didn't realize it is the primary documentation page for the language.

### DIFF
--- a/data/menus.yaml
+++ b/data/menus.yaml
@@ -24,6 +24,7 @@
   items:
     - Learning resources (start here): learn.html
     - API documentation: https://leanprover-community.github.io/mathlib4_docs
+    - Declaration search (Loogle): https://loogle.lean-lang.org/
     - Calc mode: extras/calc.html
     - Conv mode: extras/conv.html
     - Simplifier: extras/simp.html


eric-wieser:
I think the distinction that "API documentation" is trying to make is that this is "here is a short bit of text for every function" as opposed to "here is a high-level tutorial on how to use mathlib".

data/menus.yaml
```diff
@@ -23,7 +23,8 @@
 - title: Documentation
   items:
     - Learning resources (start here): learn.html
-    - API documentation: https://leanprover-community.github.io/mathlib4_docs
+    - General documentation: https://leanprover-community.github.io/mathlib4_docs
+    - Definition search (Loogle): https://loogle.lean-lang.org/
```
eric-wieser, line 27: Loogle searches theorems too, so maybe

```suggestion
    - Semantic search (Loogle): https://loogle.lean-lang.org/
```

omentic:
Hmm, that makes sense, but I still feel like the current description isn't super clear. (If #513 gets merged it's probably fine though?) I would like to call it "Stdlib documentation" but it documents so much more than just the standard library...

PR #518, opened 2024-08-22
Update Fordham course info

hrmacbeth:


data/courses.yaml
```diff
@@ -211,19 +211,21 @@
...
   experiences: >
-    The students liked using Lean and were able to learn it at the pace required.
+    The course has been taught since 2023 and is now on its fourth iteration.
+    The students like using Lean and are able to learn it at the pace required.
     I invested a lot of energy in writing custom tactics which perform exactly one step of reasoning that I would permit on paper,
     and deliberately never mentioned several more powerful tactics.
     This is very useful if the course is explicitly supposed to teach good style in proof writing.
     It lets you off the hook in grading -- you can just say "that's not enough detail -- Lean wouldn't accept it!"
-  year: 2023
+  year: 2024
```
eric-wieser, line 228: Should we have a means to record multiple years here? I assume duplicating the entry would not be desirable!

PR #520, opened 2024-08-30
Updates outdated link to lake documentation

kyleellefsen:
The old documentation pointed to the stand alone repo. This points to the corresponding file in https://github.com/leanprover/lean4


eric-wieser:
Thanks!

PR #521, opened 2024-09-05
feat: add dependency graph

joneugster:
Restore the dependency graph, which used to exist for Lean 3 and was originally added in #218.

- [x] depends on: leanprover-community/mathlib4_docs#3


joneugster:
https://leanprover-community.github.io/mathlib4_docs/mathlib.html

fpvandoorn:
Oh nice, we're now building this by default? Cool!

PR #534, opened 2024-10-01
Update the contribution guide

YaelDillies:
As pointed out in https://github.com/leanprover-community/leanprover-community.github.io/pull/359#discussion_r1312346046, `- [ ] depends on:` should be below the `---`. I think most people already place it in the correct position because the template tells them so, but we might as well update the contribution guide too.

### DIFF
--- a/templates/contribute/commit.md
+++ b/templates/contribute/commit.md
@@ -81,15 +81,17 @@ docs(CategoryTheory/EssentialImage): typo and punctuation
 Fix a typo, add two periods.
 ```
 
-an example with dependent PRs:
+An example with dependent PRs:
 
 ```markdown
 feat: The norm on `Unitization` is a C⋆-norm
 
 This shows that C⋆-algebras are always `RegularNormedAlgebra`s, so that their `Unitization` is equipped with a norm. Moreover, we show this norm is a C⋆-norm.
 
+---
+
 - [ ] depends on: #5330
 - [ ] depends on: #5741
-- [ ] depends on: #5742 
-- [ ] depends on: #5743 
+- [ ] depends on: #5742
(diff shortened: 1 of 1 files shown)


templates/contribute/commit.md
```diff
@@ -41,7 +41,7 @@ For instance, it could be
 `<subject>` has the following constraints:
 
 - use imperative, present tense: "change" not "changed" nor "changes"
-- do not capitalize the first letter
+- capitalize the first letter
```
b-mehta, line 44: I don't think this is a good idea. 

> these were about half/half upper and lower case. 

Looking at PR titles from the first half of 2024, the majority were lower case, with the causes of exceptions being mostly new contributors.

> The typography of English sentences

PR titles are not English sentences, for instance they do not allow a `.` at the end. Similarly the grammatical structure of an English sentence is also not relevant in the context of a PR title.

> The requirement for the PR description

PR descriptions have a very different purpose from PR titles, so this is not a fair comparison.

> The commit messages auto-generated by git when doing a merge ("Merged remote-tracking branch ...")

Similarly to the previous, commit messages are not prefixed by a type, and thus must follow a different convention. This is intentional, and not a reason to change either. Note that the commit messages auto-generated by bots on PRs as well as when making PRs follow the lowercase convention.

> The statu quo on other leanprover-community repos, eg this one.

Other repos do not have a style guide, whereas the status quo for the mathlib repo and the Lean repo is to follow the lowercase convention.

b-mehta: In fact, looking at the roughly 14 thousand commits to mathlib4, less than 2 thousand of these were upper-cased. Thus this particular change is a pretty significant deviation from what is done in practice.

PR #540, opened 2024-10-21
fix build by quoting line with stray colon

bryangingechen:
fallout from #536


bryangingechen:
It looks like CI didn't run on that PR automatically because it was from a fork. I've added a required check so that we remember to click the button to enable CI on such PRs.

PR #546, opened 2024-11-02
chore: add tags to all ITP24 papers

grunweg:
I verified that all are using Lean 4.
I am reasonably confident that 'formalization' is a reasonable categorisation of them all; I'm happy to change the tag for the *duper* paper.


fpvandoorn:
I just merged the other PR to get both.

PR #549, opened 2024-11-12
Added course at RWTH Aachen

PhoenixIra:
We added lean4 a bit spontaneous as part of the exercises accompanying the course Semantics and Verification of Software.


fpvandoorn:
Thanks!

PR #125, opened 2020-09-01
get 100.yaml, overview.yaml and undergrad.yaml from mathlib_docs

bryangingechen:
cf. [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/website.20overview/near/208659351), leanprover-community/mathlib#4016, and leanprover-community/doc-gen#60.

I also updated `.gitignore` and `README.md` and also moved all the `urllib.request.urlretrieve` lines next to where the downloaded files were used in `make_site.py`.

PR #558, opened 2024-11-30
feat: add page for tracking the 1000+ theorems list

grunweg:
Depends on https://github.com/leanprover-community/mathlib4_docs/pull/5.
Details of the generated webpage are up for discussion.

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -255,7 +255,7 @@ class NTheorems(Enum):
 
 def download_N_theorems(kind: NTheorems) -> dict:
     if kind == NTheorems.Hundred:
-        (fname, Type, name) = ('100.yml', HundredTheorem, 'hundred_theorems')
+        (fname, Type, name) = ('100.yaml', HundredTheorem, 'hundred_theorems')
     else:
         (fname, Type, name) = ('1000.yml', ThousandPlusTheorem, 'thousand_theorems')
     if DOWNLOAD:
--- a/make_site.py
+++ b/make_site.py
@@ -176,7 +176,9 @@ class HundredTheorem:
 
 @dataclass
 class ThousandPlusTheorem:
-    # Wikidata identifier: the letter Q followed by a string as digits, e.g. "Q1008566"
+    # Wikidata identifier (the letter Q followed by a string as digits),
+    # optionally followed by a letter (such as "A", "B" or "X" for disambiguation).
+    # "Q1008566" and "Q4724004A" are valid identifiers, for example.
     wikidata: str
     title: str
     decl: Optional[str] = None
--- a/make_site.py
+++ b/make_site.py
@@ -711,6 +711,8 @@ def render_site(target: Path, base_url: str, reloader=False, only: Optional[str]
                 ('papers.html', {'paper_lists': paper_lists}),
                 ('100.html', {'hundred_theorems': hundred_theorems}),
                 ('100-missing.html', {'hundred_theorems': hundred_theorems}),
+                ('1000.html', {'thousand_theorems': thousand_theorems}),
+                ('1000-missing.html', {'thousand_theorems': thousand_theorems}),
                 ('meet.html', {'users': users,
                                'community': read_md('community.md')
                                }),
--- a/make_site.py
+++ b/make_site.py
@@ -176,7 +176,7 @@ class HundredTheorem:
 
 @dataclass
 class ThousandPlusTheorem:
-    # Wikidata identifier, e.g. "Q1008566"
+    # Wikidata identifier: the letter Q followed by a string as digits, e.g. "Q1008566"
     wikidata: str
     title: str
     decl: Optional[str] = None
@@ -257,7 +257,7 @@ def download_N_theorems(kind: NTheorems) -> dict:
     if kind == NTheorems.Hundred:
         (fname, Type, name) = ('100.yaml', HundredTheorem, 'hundred_theorems')
     else:
-        (fname, Type, name) = ('1000.yml', ThousandPlusTheorem, 'thousand_theorems')
+        (fname, Type, name) = ('1000.yaml', ThousandPlusTheorem, 'thousand_theorems')
     if DOWNLOAD:
         download(f'https://leanprover-community.github.io/mathlib4_docs/{fname}', DATA/fname)
         with (DATA/fname).open('r', encoding='utf-8') as h_file:
--- a/make_site.py
+++ b/make_site.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from enum import Enum, auto
 from pathlib import Path
 import sys
 import subprocess
@@ -176,7 +177,7 @@ class HundredTheorem:
 @dataclass
 class ThousandPlusTheorem:
     # Wikidata identifier, e.g. "Q1008566"
-    number: str
+    wikidata: str
     title: str
     decl: Optional[str] = None
     decls: Optional[List[str]] = None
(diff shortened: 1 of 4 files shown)


templates/1000.html
```diff
@@ -0,0 +1,23 @@
+{% extends '_base.html' %}
+{% block title %}100 theorems in Lean{% endblock %}
```
fpvandoorn, line 2:
```suggestion
{% block title %}1000+ theorems in Lean{% endblock %}
```

make_site.py
```diff
@@ -236,39 +248,40 @@ class InfoEntry:
...
+                if h.decl:
+                    assert not h.decls
+                    h.decls = [h.decl]
+                if h.decls:
+                    doc_decls = []
+                    for decl in h.decls:
+                        try:
+                            decl_info = declarations[decl]
+                        except KeyError:
+                            print(f'Error: 100 theorems entry {h.number} refers to a nonexistent declaration {decl}')
```
fpvandoorn, line 266: hard-coded 100

make_site.py
```diff
@@ -173,6 +173,18 @@ class HundredTheorem:
     links: Optional[Mapping[str, str]] = None
     note: Optional[str] = None
 
+@dataclass
+class ThousandPlusTheorem:
+    # Wikidata identifier, e.g. "Q1008566"
+    number: str
```
fpvandoorn, line 179: the "number" occasionally ends in an `X`/`A`/`B`, which is probably included in this field. Maybe add a comment that this does include these trailing letters?

fpvandoorn (approved):
Mostly LGTM

templates/1000.html
```diff
@@ -0,0 +1,23 @@
+{% extends '_base.html' %}
+{% block title %}1000+ theorems in Lean{% endblock %}
+{% block content %}
+
+        <h1>1000+ theorems</h1>
+
+        <p><a href="https://www.cs.ru.nl/~freek/">Freek Wiedijk</a> started the <a href="https://1000-plus.github.io/">1000+ theorems project</a> tracking progress of theorem provers in formalizing named theorems on wikipedia, as a way of comparing prominent theorem provers.
+        Currently {{thousand_theorems|selectattr('author')|list|length}} of them are formalized in Lean.
+        We also have a page with the theorems from the list <a href="1000-missing.html">not yet in Lean</a>.</p>
```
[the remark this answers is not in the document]
grunweg: Good point; I'll try to add something.

grunweg:
By the way: was one of you able to test this locally? I'm trying to install the necessary dependencies, but `pip install -r requirements.txt` doesn't feel like cooperating today: I get connection time-outs very frequently, now got to conflicting package version requirements (so am trying again in a completely new virtualenv, with all the wait this entails)... long story short, if you already have everything set up, you'll be able to test much faster than I can.

templates/1000.html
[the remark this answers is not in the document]
grunweg: The overview page had a suitable paragraph, which I have just copied here.

bryangingechen:
I haven't tested it locally yet, but I'll try and do so later today.

make_site.py
```diff
@@ -236,39 +249,51 @@ class InfoEntry:
...
-    pkl_dump('hundred_theorems', hundred_theorems)
-else:
-     hundred_theorems = pkl_load('hundred_theorems', dict())
+class NTheorems(Enum):
+    Hundred = auto()
+    ThousandPlus = auto()
 
+def download_N_theorems(kind: NTheorems) -> dict:
+    if kind == NTheorems.Hundred:
+        (fname, Type, name) = ('100.yml', HundredTheorem, 'hundred_theorems')
```
bryangingechen, line 258:
```suggestion
        (fname, Type, name) = ('100.yaml', HundredTheorem, 'hundred_theorems')
```

make_site.py
```diff
@@ -236,39 +249,51 @@ class InfoEntry:
...
+    else:
+        (fname, Type, name) = ('1000.yml', ThousandPlusTheorem, 'thousand_theorems')
```
bryangingechen, line 260: Both https://leanprover-community.github.io/mathlib4_docs/1000.yml or https://leanprover-community.github.io/mathlib4_docs/1000.yaml are 404 errors - is there a PR to doc-gen this depends on? (Sorry if I missed it somewhere above)

fpvandoorn: @grunweg It seems easier to just read the data from https://raw.githubusercontent.com/leanprover-community/mathlib4/refs/heads/master/docs/100.yaml and https://raw.githubusercontent.com/leanprover-community/mathlib4/refs/heads/master/docs/1000.yaml ?

grunweg: @bryangingechen Good point. I checked; this needs a PR to doc-gen, here it is (almost trivial diff):
https://github.com/leanprover-community/mathlib4_docs/pull/5

grunweg: @fpvandoorn Good question. I'm not sure which way is best. I like the conceptual separation of "docs builds mathlib and makes those overview files available, the webpage only reads the docs and not the mathlib repo". Conceptually, these overview files feel similar to generated documentation to me (both are externally useful artifacts, derived from the mathlib source in some sense --- one could imagine generating 100.yaml automatically from a `100theorems` attribute, for instance).
In other words: the current system makes some sense to me, and I'm not sure if there's a *good* reason to change it. "You need to just through another hoop" could be one; perhaps this just needs to be documented better?

In any case, I'd want the handling for `{undergrad,overview}.yaml` to be consistent with this --- so I'd want to change two more places in the code. That's not a big issue though.

grunweg: I'm happy to discuss which of these locations makes the most sense. But perhaps that should be a broader discussion/I'd like to have at least one more opinion here.

bryangingechen: Here's the change where we first started reading data files from mathlib_docs: https://github.com/leanprover-community/leanprover-community.github.io/pull/125

And [here's the relevant Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/website.20overview/near/208659351), where @robertylewis proposes the idea of hosting the yaml files with doc-gen. I'm guessing the idea of linking directly to mathlib files at raw.githubusercontent.com just didn't occur to us at the time. Using the raw.githubusercontent.com links has the advantage of making things more direct, although as you say, maybe someone external is already depending on the files at mathlib4_docs.

fpvandoorn: Ok, that's fine with me as well.

I'm happy with this PR if Bryan is.

make_site.py
```diff
@@ -173,6 +174,18 @@ class HundredTheorem:
     links: Optional[Mapping[str, str]] = None
     note: Optional[str] = None
 
+@dataclass
+class ThousandPlusTheorem:
+    # Wikidata identifier: the letter Q followed by a string as digits, e.g. "Q1008566"
+    wikidata: str
```
fpvandoorn, line 180: One thing, just because to be precise: this is the identifier *with* an additional `X`/`A`/`B` appended, right? So this is not technically the Wikidata identifier, but a 1000+-theorems identifier. Could you be specific in the comment at least?

make_site.py
bryangingechen: Yes, I'm happy with this as is now!

make_site.py
grunweg: Good catch, fixed!

make_site.py
grunweg: Thanks for digging up the context. I've read the zulip thread and the PRs now. If I understand correctly, there is a synchronisation constraint: it would be nice to keep the declarations (generated from doc-gen) and the web-page in sync --- so retrieving both from doc-gen4 seems reasonable.
I don't know how often doc-gen fails these days - but if it did, having a webpage using the old data consistently (rather than some weird mixture, and having to think about all corner cases this can create) strikes me as a better trade-off. If failing docs are an issue, setting up an alarm on failures seems better.

TL;DR: I think the current design is better than the other version, and will stick to it.

grunweg:
For future readers: [this comment thread](https://github.com/leanprover-community/leanprover-community.github.io/pull/558#discussion_r1889985245) discusses why these files are retrieved from doc-gen, and *not* just from raw.githubusercontent.com.

grunweg:
So, all in all: I think this PR is ready to be merged. Thanks for the careful review!

bryangingechen:
@grunweg Looks like the build is failing due to an entry in 1000.yaml with tags that are not defined in the `ThousandPlusTheorem` class: https://github.com/leanprover-community/leanprover-community.github.io/actions/runs/12402292596/job/34623513596#step:6:35

Cf. https://github.com/leanprover-community/mathlib4/blob/master/docs/1000.yaml#L143

Does it make sense to fix the YAML file / add more checks on the mathlib side, or be more flexible over here?

bryangingechen:
(In the interest of getting the website build working again ASAP, I've opted for the latter for now in #559)

grunweg:
Sorry for that; thanks for the quick fix. I'll look at this later today.

grunweg:
Proper fix at #560; needs mathlib4#20066 to be merged.

PR #561, opened 2024-12-19
remove duplicated field from `ThousandPlusTheorem` dataclass, add comments refering to `scripts/yaml_check.py` in mathlib4

bryangingechen:
cf. https://github.com/leanprover-community/mathlib4/pull/20078

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -156,6 +156,7 @@ with (DATA/'teams.yaml').open('r', encoding='utf-8') as t_file:
                   use_biography=team.get('use_biography', True))
              for team in yaml.safe_load(t_file)]
 
+# keep in sync with yaml_check.py in mathlib4 repo
 @dataclass
 class DocDecl:
     name: str
@@ -163,6 +164,7 @@ class DocDecl:
     docs_link: str
     src_link: str
 
+# keep in sync with yaml_check.py in mathlib4 repo
 @dataclass
 class HundredTheorem:
     number: str
@@ -174,6 +176,7 @@ class HundredTheorem:
     links: Optional[Mapping[str, str]] = None
     note: Optional[str] = None
 
+# keep in sync with yaml_check.py in mathlib4 repo
 @dataclass
 class ThousandPlusTheorem:
     # Wikidata identifier (the letter Q followed by a string as digits),
--- a/make_site.py
+++ b/make_site.py
@@ -191,8 +191,6 @@ class ThousandPlusTheorem:
     date: Optional[str] = None
     url: Optional[str] = None
     note: Optional[str] = None
-    url: Optional[str] = None
-
 
 @dataclass
 class Event:


grunweg (commented):
Assuming the mathlib PR is merged as-is, this PR looks good to me.
I'm not too happy with copy-pasting the dataclass declarations, but I don't see a better solution either.

fpvandoorn (approved):

grunweg:
I've created #562 adding some documentation to the DocDecl fields. (I couldn't push easily to this branch, so submitted that as a separate PR. Feel free to copy-paste some of that to the mathlib4 PR; I'm checking out for tonight.)

bryangingechen:
@grunweg How about this? https://github.com/leanprover-community/leanprover-community.github.io/pull/561/commits/da64b74f3bde9aced1d9eb08c49dd4aac56ae9f3

After this refactor, `HundredTheorem` has an "index" field `number` and `ThousandPlusTheorem` has `wikidata`, but otherwise they share the same set of "public" fields:
```python
    title: str
    decls: Optional[List[str]] = None
    author: Optional[str] = None
    date: Optional[str] = None
    links: Optional[Mapping[str, str]] = None
    note: Optional[str] = None
```
Regarding "public" fields, here's the idea:
- In `make_site.py`, the datatype definition also includes a private field `_doc_decls`, (I just renamed `doc_decls` to make it more obvious that it's meant to be private).
- In `check_yaml.py` in https://github.com/leanprover-community/mathlib4/pull/20078, we will only include the public fields, and will NOT include `_doc_decls`, so that the check there will not allow that field to be included.

This feels a bit hacky, but I think it beats trying to squeeze `decls` and `doc_decls` into the same field (since the template code is already calling the various parts of `doc_decls`) or having public and private versions of the full datatypes.

bryangingechen:
We could also combine `HundredTheorem` and `ThousandPlusTheorem` into a single class too if we used an `id` field rather than `number` + `wikidata`. Then the only question is would be what to name that class...

grunweg:
Thanks for asking. Hm, I'm not sure. In any case, *meant* to say something else:
- I was proposing to create a new class "TheoremForWebpage" or so; that class would contain `doc_decls` and only a `decls` field
- then `{Hundred,ThousandPlus}Theorem` could remove `doc_decls`, have just an `id` field etc, while the yaml files could stay mostly as-is

grunweg:
I've just sat down: #563 is approximately what I would change instead. Let me know what you think!

make_site.py
```diff
@@ -170,36 +172,40 @@ class DocDecl:
     # is simply a link to the right revision of the mathlib source code.
     src_link: str
 
+# keep public fields in sync with `scripts/yaml_check.py` in mathlib4 repo
 @dataclass
 class HundredTheorem:
     number: str
     title: str
-    decl: Optional[str] = None
```
grunweg, line 177: Removing this requires changing 100.yaml, imho for the worse. I presume this structure with `decl` and `decls` is to allow nicer syntax for the common case of just a single declaration, which this change would break.

make_site.py
```diff
@@ -170,36 +172,40 @@ class DocDecl:
...
     decls: Optional[List[str]] = None
-    # The HTML source code for the generated documentation entry
-    # for the declaration(s) in |decl| or |decls|.
-    doc_decls: Optional[List[DocDecl]] = None
     author: Optional[str] = None
     date: Optional[str] = None
-    url: Optional[str] = None
+    links: Optional[Mapping[str, str]] = None
```
grunweg, line 202: To be honest, I like the `url` field for `1000.yaml`. (This also matches the upstream specification, which just allows putting one URL.) I'd rather not change this aspect.
You *are* right that 1000.html was previously wrong, but I'd propose to unify these conventions at a different level.

make_site.py
bryangingechen: Yes, I agree.

make_site.py
bryangingechen: I'm actually a bit surprised that you prefer `url`, given that there's already an entry in `1000.yaml` where it seems having a `links`-style list of pairs would be more flexible: https://github.com/leanprover-community/mathlib4/blob/62fd5798b4a61e5cbeee6e32562dcb4cbd66e0b1/docs/1000.yaml#L145

However, as I said above, I'm happy to go along with you on this.

bryangingechen:
Thanks for making #563; my first impression is that if you're happy with it then we can close this PR and just work over there, but let me take a more detailed look later today.

make_site.py
grunweg: If we have a single `url` field upstream, but allow `links` here, there will be pain converting between these two. I'm not convinced that is a good strategy. That said: you *are* right that Gödel's incompleteness theorem should be split in two. In other words, one should file a PR upstream, splitting this into two results. I'll add that to my TODO list.

grunweg: @fpvandoorn What do you think about this question? Should one have a single URL upstream, or allow multiple URLs? (Case in point, Gödel's incompleteness theorem.) I think having a single URL is nice --- and would thus propose splitting the upstream entry instead. What do you think?

(Not urgent: we can discuss this in the new year.)

grunweg:
I just polished the description of #563: from my side, I'm happy to merge that one, if you're also happy with it.

PR #562, opened 2024-12-19
doc/chore: document `docDecls` field

grunweg:
As well as a hack with the `header-data.json` file.

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -176,6 +176,8 @@ class HundredTheorem:
     title: str
     decl: Optional[str] = None
     decls: Optional[List[str]] = None
+    # The HTML source code for the generated documentation entry
+    # for the declaration(s) in |decl| or |decls|.
     doc_decls: Optional[List[DocDecl]] = None
     author: Optional[str] = None
     links: Optional[Mapping[str, str]] = None
@@ -190,6 +192,8 @@ class ThousandPlusTheorem:
     title: str
     decl: Optional[str] = None
     decls: Optional[List[str]] = None
+    # The HTML source code for the generated documentation entry
+    # for the declaration(s) in |decl| or |decls|.
     doc_decls: Optional[List[DocDecl]] = None
     author: Optional[str] = None
     date: Optional[str] = None
--- a/make_site.py
+++ b/make_site.py
@@ -156,11 +156,18 @@ with (DATA/'teams.yaml').open('r', encoding='utf-8') as t_file:
                   use_biography=team.get('use_biography', True))
              for team in yaml.safe_load(t_file)]
 
+# Data for a documentation entry for a single declaration.
 @dataclass
 class DocDecl:
+    # Fully qualified name of the declaration
     name: str
+    # Full HTML code for this declaration's entry on its generated documentation page.
     decl_header_html: str
+    # URL of this declaration's entry in the generated documentation.
+    # This is site-relative, and starts with "/mathlib4_docs/".
     docs_link: str
+    # URL for this declaration's generated source entry: currently,
+    # is simply a link to the right revision of the mathlib source code.
     src_link: str
 
 @dataclass
--- a/make_site.py
+++ b/make_site.py
@@ -218,6 +218,11 @@ class Course:
 
 if DOWNLOAD:
     print('Downloading header-data.json...') #  This is a slow operation, so let's inform readers.
+    # header-data.json contains information for every single declaration in mathlib
+    # which has a generated documentation entry (e.g., skipping private declarations by default).
+    # The resulting file is huge (several hundred MB), hence we use a small HACK:
+    # doc-gen4 uploads this file as a .bmp file, so GitHub's servers will serve it in
+    # compressed form. When downloading it locally, we save it (decompressed) with the correct extension.
     download(
         'https://leanprover-community.github.io/mathlib4_docs/declarations/header-data.bmp',
         DATA/'header-data.json'


grunweg:
I'm not quite sure why the bmp/json hack was created. Help reconstructing this and expanding/fixing the comment is welcome.

make_site.py
```diff
@@ -218,6 +226,13 @@ class Course:
 
 if DOWNLOAD:
     print('Downloading header-data.json...') #  This is a slow operation, so let's inform readers.
+    # header-data.json contains information for every single declaration in mathlib
+    # which has a generated documentation entry (e.g., skipping private declarations by default).
+    # The resulting file is huge (several hundred MB), hence we use a small HACK:
+    # doc-gen4 uploads this file as a .bmp file, so browsers will download it compressed,
+    # and employ their machinery for dealing with large files. XXX is this true at all?
+    # When downloading it locally, we save it with the correct extension.
+    # why? blocklisting of file extensions? automatic compression in web browsers?
```
bryangingechen, line 235:
```suggestion
    # doc-gen4 uploads this file as a .bmp file, so GitHub's servers will serve it in
    # compressed form. When downloading it locally, we save it with the correct extension, 
    # so that we can more easily inspect the file locally for debugging.
```
Here are my best guesses to the answers to your questions.

grunweg: That looks very plausible - reading the linked discussion, that seems to be it. I've tweaked this slightly and adopted it. Thanks!

grunweg:
I have also documented the `doc_decl` fields. From my side, this PR is good to go.

After merging this PR, one could
- add the same docs in mathlib
- synchronise any comments on the mathlib side here
- merge everything :-)

bryangingechen:
After reading through the code again, and doing some archaeology to when the `doc_decls` field was first added in https://github.com/leanprover-community/leanprover-community.github.io/pull/54, I now think `doc_decls` was never meant to be exposed on the YAML side and is more of an internal implementation detail (perhaps this was obvious to you the whole time).

Importantly, anything contained in the `doc_decls` field coming from the YAML file seems to be overwritten in [these lines](https://github.com/leanprover-community/leanprover-community.github.io/pull/562/files#diff-0bafb17275f3075935a3b5576351f24231b1c5d964d5099b77260c3c2cc416d9R286-R311). 

The "right" thing to do here might be to refactor the code so that we have one datatype for the object parsed from the YAML and one datatype for the object that eventually gets passed to the HTML template, but I'm also getting the feeling that we're already over-engineering this, so maybe the easiest thing for now is just to note (in various places) the fact that `doc_decls` exists in the type for technical reasons, but shouldn't be used in the YAML file.

I could be missing something, so happy to hear everyone's thoughts!

grunweg:
Thanks for doing the archaeology! 

I agree: the `doc_decls` field should *never* be present in the yaml files in mathlib4 (and it would be good to enforce this). This is generated only for displaying the template page.

One could simplify the template entries, by
- having just a single list of declaration names (merging decl and decls, like the code does anyway),
- merging decl(s) and doc_decls into one `List[str, DocDecl]` (or perhaps `List[str, DocDecl | None]`)
- merging `links` and `url` into one list(?), so there'd be only one data type for these templates
This would make the code slightly better-typed, but also longer.

I am equally fine with enforcing the separation in `yaml-check.py` and with having *one* extra type.


----------

Side note/not the main point: I don't agree with your phrasing in
> Importantly, anything contained in the doc_decls field coming from the YAML file seems to be overwritten in [these lines](https://github.com/leanprover-community/leanprover-community.github.io/pull/562/files#diff-0bafb17275f3075935a3b5576351f24231b1c5d964d5099b77260c3c2cc416d9R286-R311).

(Perhaps that's me over-interpreting what you say.) I wrote the following, which I'll copy here just in case. As I understand it,
- the 100 theorems page (and the 1000+ theorems page by copying from it) uses the `doc_decls` field to print the statements of the formalised theorems. (Indeed, they seemingly use literally the same HTML code, just with different styling.) A clever trick/small hack.
- the lines you link overwrite the `doc_decls` field, but the main content still comes from `header-data.json`
Let me walk through that explicitly
- `declarations` is populated a few lines above by reading `header-data.json`
- the loop makes a local copy of the `HeaderDataEntry` for each declaration referenced for the theorem
- that data is modified slightly, to add the `mathlib4_docs` path prefix (the .json file uses doc-relative links, which are converted to site-relative)
- apart from that, the data is passed through: the source code for `header` virtually comes from `header-data.json`

bryangingechen:
I think we're largely in agreement here. I probably phrased the "overwritten" part of my comment poorly; I just meant that anything that may have been provided in the `doc_decls` field in `100.yaml` or `1000.yaml` is replaced, as you say, with data from `header-data.json`. 

I'm also OK with making changes to the template format - since it looks like you and @fpvandoorn will be doing a lot of work with `1000.yaml`, I'm happy to go with whatever will make your lives easier now and in the future. (Probably better to do all this now before we forget all about this and end up having to do archaeology again...)

bryangingechen:
@grunweg did you want to take the first stab at defining a new datatype for the template? If not, I'll merge this and then give it a shot myself in #561 and the associated mathlib PR.

grunweg:
I won't have too much time to work on this - feel free to go ahead!

grunweg:
I still guess what I sketched could work. If you find out it doesn't, then go by a better design, of course :-)

PR #566, opened 2025-01-03
extra `calc` style

YaelDillies:
I find this style pretty convenient, as it has vertical alignment AND ease of editing.

### DIFF
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -363,8 +363,25 @@ theorem reverse_reverse : ∀ (l : List α), reverse (reverse l) = l
       _ = a :: l := rfl
 ```
 
-However, because the expressions and proofs are relatively short, the following style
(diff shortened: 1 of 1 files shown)


fpvandoorn:
You insert an extra underscore there? I didn't know that.
Did you discuss on Zulip to add this to the style? Seems reasonable to me, but it is also pretty confusing (it seems like you're applying `reverse (reverse (a :: l))` to `_`)

YaelDillies:
It was discussed on Zulip [here](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/calc.20indentation.3A.20style.20guide.3F/near/422234173).

> it seems like you're applying `reverse (reverse (a :: l))` to `_`

If you indent the LHS more, I doubt this confusion can really arise.

PR #571, opened 2025-01-09
Run the build on pull requests as well

Ruben-VandeVelde:
`deploy.sh` already has a `if [ "$github_repo" = "leanprover-community/leanprover-community.github.io" -a "$github_ref" = "refs/heads/lean4" ]; then` check that should make sure the site is only built, not deployed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
     branches-ignore:
       - 'master'
       - 'oldsite'
+  pull_request:
   schedule:
     - cron: '0 4 * * *'
   workflow_dispatch:


.github/workflows/build.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches-ignore:
       - 'master'
       - 'oldsite'
+  pull_request:
```
bryangingechen, line 8: The build process relies on some GitHub API tokens in the repo secrets that PRs from forks don't usually have access to, but maybe [`pull_request_target`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target) will work?

```suggestion
  pull_request_target:
```

PR #574, opened 2025-01-09
change logic for which theorems count as formalized

bryangingechen:
(and move out of templates into make_site.py)

cf. [Zulip](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/1000-plus.20theorems.20project/near/492814764)

### DIFF
--- a/make_site.py
+++ b/make_site.py
@@ -255,6 +255,8 @@ class TheoremForWebpage:
     author: Optional[str] = None
     date: Optional[str] = None
     note: Optional[str] = None
+    # A boolean tagging theorems that have been formalized
+    formalized: bool
 
 
 @dataclass
@@ -362,7 +364,9 @@ def download_N_theorems(kind: NTheorems) -> dict:
                             # note: the `header-data.json` data file uses doc-relative links
                             docs_link='/mathlib4_docs/' + decl_info.info.docLink,
                             src_link=decl_info.info.sourceLink))
-                theorems.append(TheoremForWebpage(id, h.title, doc_decls, links, h.author, h.date, note))
+                # a theorem counts as formalized if the author field is nonempty or if doc_decls is nonempty
+                formalized = bool(h.author) or (len(doc_decls) > 0)
+                theorems.append(TheoremForWebpage(id, h.title, doc_decls, links, h.author, h.date, note, formalized))
         pkl_dump(name, theorems)
     else:
         theorems = pkl_load(name, dict())
--- a/make_site.py
+++ b/make_site.py
@@ -364,7 +364,7 @@ def download_N_theorems(kind: NTheorems) -> dict:
                             # note: the `header-data.json` data file uses doc-relative links
                             docs_link='/mathlib4_docs/' + decl_info.info.docLink,
                             src_link=decl_info.info.sourceLink))
-                # a theorem counts as formalized if the author field is nonempty or if doc_decls is nonempty
+                # A theorem counts as formalized if the author field or `doc_decls` is non-empty.
                 formalized = bool(h.author) or (len(doc_decls) > 0)
                 theorems.append(TheoremForWebpage(id, h.title, formalized, doc_decls, links, h.author, h.date, note))
         pkl_dump(name, theorems)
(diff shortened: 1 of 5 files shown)


templates/100.html
```diff
@@ -16,9 +16,9 @@ <h1>100 theorems</h1>
 	<a href="https://docs.github.com/en/github/managing-files-in-a-repository/editing-files-in-another-users-repository">"Editing files in another user's repository"</a>.
 	</p>
 
-    {% for theorem in hundred_theorems|selectattr('author') %}
+    {% for theorem in hundred_theorems|selectattr('formalized') %}
         <h5 class="markdown-heading mt-5" id="{{ theorem.id }}">{{ theorem.id }}. {{ theorem.title }} <a class="hover-link" href="#{{ theorem.id }}">#</a></h5>
-        <p>Author{% if ' and ' in theorem.author %}s{% endif %}: {{ theorem.author }}</p>
+        {% if theorem.author %}<p>Author{% if ' and ' in theorem.author %}s{% endif %}: {{ theorem.author }}</p>{% endif %}
```
bryangingechen, line 21: If there's no author data, I chose to just skip this line, but we could also say something like "Author data incomplete" or something like that to spur people into filling in the data.

make_site.py
```diff
@@ -362,7 +364,9 @@ def download_N_theorems(kind: NTheorems) -> dict:
                             # note: the `header-data.json` data file uses doc-relative links
                             docs_link='/mathlib4_docs/' + decl_info.info.docLink,
                             src_link=decl_info.info.sourceLink))
-                theorems.append(TheoremForWebpage(id, h.title, doc_decls, links, h.author, h.date, note))
+                # a theorem counts as formalized if the author field is nonempty or if doc_decls is nonempty
```
grunweg, line 367:
```suggestion
                # A theorem counts as formalized if the author field or `doc_decls` is non-empty.

File: templates/100.html (hunk 1 above)
Comment (grunweg) [reply templates/100.html:21]: Good thinking. I think your version is fine.

File: templates/1000-missing.html [hunk 3]
```diff
@@ -4,8 +4,9 @@
 {% markdown %}
 # Missing theorems from [Freek Wiedijk's](https://www.cs.ru.nl/~freek/) [1000+ theorems project](https://1000-plus.github.io/)
 These theorems are not yet formalized in Lean (or, these formalisations are not entered in the database yet).
+Currently there are {{hundred_theorems|rejectattr('formalized')|list|length}} of them.
```
grunweg, line 7: I like this addition, but careful: "hundred_theorems" should be updated, right?

grunweg (commented):
Nice and simple, I like it. (I was thinking of combining two `rejectattr` for the `-missing.html` pages, but that wouldn't have worked for the other two.)
One comment tweak and one copy-paste error, I think: otherwise, LGTM.
maintainer delegate (I know that doesn't work here)

templates/1000-missing.html
bryangingechen:
```suggestion
Currently there are {{thousand_theorems|rejectattr('formalized')|list|length}} of them.
```

bryangingechen:
@fpvandoorn I think this one is ready to go now too.

fpvandoorn:
This looks good to me.

There is one axis along we could further future-proof this: the `1000-plus` repo has a field to indicate whether a theorem is just stated formally, but not necessarily proven.
> `status`: `formalized` (the proof is formalized), `statement` (just the statement is)

This makes most sense for theorems that are very hard to even state. It might be nice to support that as well on our side, which presumably requires adding an additional field, and then the logic here can be simplified to just read the entries of that field.

But I'm also fine to decide later whether someone will actually want to use this ability, and keep this PR like this for now, so feel free to merge this as-is.

grunweg:
https://github.com/leanprover-community/mathlib4/pull/20637 shows the mathlib changes required; I have the easy half of the webpage prepared locally. I'd like this PR to land before polishing those (if at all).

bryangingechen:
OK, let's merge this first then. Thanks for the reviews!

PR #575, opened 2025-01-10
Avoid including invalid links in 1000.html

Ruben-VandeVelde:
The current code includes `<p><a href="None">url</a></p>` for every entry that doesn't have a URL listed.


bryangingechen:
Tested locally and it works, thanks!

PR #576, opened 2025-01-10
1000.html: link IDs to their wikidata page

Ruben-VandeVelde:


bryangingechen:
Nice idea! Could you add similar code to `1000-missing.html` for completeness?

bryangingechen:
Thanks!

PR #578, opened 2025-01-20
chore: use 'authors' for the authors field in {Hundred,Thousand}Theorems

grunweg:
This matches the `Formalization` type in `make_site.py`, and the mathlib PR https://github.com/leanprover-community/mathlib4/pull/20875.


grunweg:
Should be merged at the same time as the mathlib PR; otherwise this webpage will break.

fpvandoorn:
The website is probably broken until the mentioned Mathlib PR lands

PR #579, opened 2025-01-20
chore(100,1000.html): remove mention of formalised statements

grunweg:
The previous PR broke the webpage build; I don't understand why. Hopefully, reverting this fixes it.


grunweg:
Depends on #578 (as these will conflict).

bryangingechen:
I get the following message, but the build completes without failing, so I'll merge this for now:
> Error: 1000+ theorems entry Q776578 refers to a nonexistent declaration isSemisimpleRing_iff_pi_matrix_divisionRing

PR #592, opened 2025-02-16
feat: add naming convention for Prop-valued classes

kbuzzard:
This is a proposal for a naming convention for Prop-valued classes. The proposal is that nouns get `Is` and adjectives do not.


kbuzzard:
I personally would prefer that every Prop-valued class gets an Is for consistency but this proposal has never seemed to be a popular one. However `TopologicalRing` has always been a terrible name for a Prop, in my opinion, regardless of anyone's thoughts on a general convention.

fpvandoorn:
LGTM. Is there a consensus on Zulip that this is what we want?

PatrickMassot:
I’m merging this after noting that https://github.com/leanprover-community/mathlib4/pull/21472 got merged.

PR #593, opened 2025-02-18
feat: add language reference to documentation overview

david-christiansen:
I think it might help users find the reference manual if it's included in the overview of documentation on this site.

### DIFF
--- a/data/menus.yaml
+++ b/data/menus.yaml
@@ -21,6 +21,7 @@
     - Learning resources (start here): learn.html
     - API documentation: https://leanprover-community.github.io/mathlib4_docs
     - Declaration search (Loogle): https://loogle.lean-lang.org/
+    - Language reference: https://lean-lang.org/doc/reference/latest/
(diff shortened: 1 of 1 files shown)


david-christiansen:
Building the site with my system Python isn't working, because one of its dependencies depends on `imp`, which was removed in Python 3.12. A PR to fix it [was closed](https://github.com/gorakhargosh/pathtools/issues/13), so I haven't managed to test this change.

bryangingechen:
Yeah, I have my version of python for this project set to 3.11.11. I've been meaning to add instructions for using `pyenv` to the `README` but haven't gotten around to it.

I'm happy with this as is, but while you're here, maybe you can suggest something to add to our [learning resources page](https://leanprover-community.github.io/learn.html), which also doesn't mention the language reference yet: https://github.com/leanprover-community/leanprover-community.github.io/blob/lean4/templates/learn.md ?

bryangingechen:
I've created #599 for the learning resources page.

bryangingechen:
Apologies, I missed that there was a mention of the language reference under the Books section already!

PR #594, opened 2025-02-19
Reorder words in sets.md

ericluap:
Just a tiny change for clarity, matches the order of the descriptions to that of the functions.


bryangingechen:
Thank you!

PR #596, opened 2025-02-19
doc(Style): clarify indentation of documentation strings

grunweg:
I searched for guidelines here, and could not find them. [Zulip](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Style.20.3Abicycle.3A.20.3A.20indenting.20second.20lines.20in.20doc-strings) yields very clear rules; let's codify them.

### DIFF
--- a/templates/contribute/style.md
+++ b/templates/contribute/style.md
@@ -80,7 +80,8 @@ two main concepts in the theory of xyzzyology.
 ## Main results
 
 - `exists_foo`: the main existence theorem of `foo`s.
-- `bar_of_foo`: a construction of a `bar`, given a `foo`.
+- `bar_of_foo`: a construction of a `bar`, given a `foo` under very complicated
+  and intricate circumstances worth describing in excruciating detail.
 - `bar_eq`    : the main classification theorem of `bar`s.
 
 ## Notation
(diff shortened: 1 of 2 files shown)


templates/contribute/style.md
```diff
@@ -80,7 +80,8 @@ two main concepts in the theory of xyzzyology.
 ## Main results
 
 - `exists_foo`: the main existence theorem of `foo`s.
-- `bar_of_foo`: a construction of a `bar`, given a `foo`.
+- `bar_of_foo`: a construction of a `bar`, given a `foo` under very complicated
+  and intricate circumstances worth describing in excruciating detail.
```
PatrickMassot, line 84: I don’t understand what this change is doing here.

grunweg: The change is meant to illustrate an example, where an item spans multiple lines. (I got this wrong previously, by not indenting the second line: this is not how markdown expects items. Hence, I think this is worth clarifying.)

It seems the wording is not clear; I'm very happy to change it. For instance, would this be clearer?
```
- `bar_of_foo_of_baz`: a construction of a `bar`, given a `foo` and a `baz`. If this doc-string was longer than one
  line, subsequent lines should be indented by two spaces (as required by markdown syntax).

grunweg: I have updated the phrasing. @PatrickMassot Does that address your comment?

PR #597, opened 2025-02-21
added Grad Course in Lyon

faenuccio:


robertylewis:
If the course and materials are in English, do you think it's worth translating the title? Personally I'd probably skip over it in the list assuming the contents are in French.

faenuccio:
> If the course and materials are in English, do you think it's worth translating the title? Personally I'd probably skip over it in the list assuming the contents are in French.

Good point, fixing it.

PR #604, opened 2025-03-10
Style: Add Deprecation sub-section

vlad902:
Zulip thread here:
https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Style.20guide.3A.20deprecation


templates/contribute/style.md
```diff
@@ -565,3 +565,27 @@ Documentation strings for declarations are delimited with `/-- -/`.
...
+
+@[deprecated "This theorem is deprecated in favor of using ... with ..." (since := "YYYY-MM-DD")]
+theorem example_thm ...
+```
+
+The [`deprecate to`](/mathlib4_docs/Mathlib/Tactic/DeprecateTo.html) command can help generate
+alias definitions. The `@[deprecated]` attribute should always include the deprecation date,
+and either the new name for the theorem when it is being renamed or kept substantially the same,
+or a string to explain how transition away from the old definition when a new version is no longer
+being provided.
```
jcommelin, line 591: This is maybe also a good place to point to `scripts/add_deprecations.sh`.

jcommelin (commented):
Thanks for documenting this!

templates/contribute/style.md
```diff
@@ -565,3 +565,27 @@ Documentation strings for declarations are delimited with `/-- -/`.
 
 See our [documentation requirements](doc.html) for more suggestions
 and examples.
+
+### Deprecation
+
+Deleting, renaming, or changing theorem definitions can cause downstreams projects
+that rely on these definitions to fail to compile. Any publicly exposed theorem
+definitions that are being changed should be gracefully transitioned by keeping the
```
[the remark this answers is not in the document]
vlad902: That's a good point. I've updated the wording to say:

> Any publicly exposed theorems and definitions that are being removed should be ...

templates/contribute/style.md
vlad902: Done.

templates/contribute/style.md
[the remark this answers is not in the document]
jcommelin: I don't see those changes...
```suggestion
that rely on these definitions to fail to compile. Any publicly exposed theorem and
definitions that are being removed should be gracefully transitioned by keeping the
```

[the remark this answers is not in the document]
vlad902: Didn't notice that git push failed, fixed.

templates/contribute/style.md
```diff
@@ -565,3 +565,29 @@ Documentation strings for declarations are delimited with `/-- -/`.
 
 See our [documentation requirements](doc.html) for more suggestions
 and examples.
+
+### Deprecation
+
+Deleting, renaming, or changing declarations can cause downstreams projects
+that rely on these definitions to fail to compile. Any publicly exposed theorems and
+definitions that are being removed should be gracefully transitioned by keeping the
+old theorem statement with a `@[deprecated]` attribute, to warn downstream projects to
```
grunweg, line 574:
```suggestion
old declaration with a `@[deprecated]` attribute, to warn downstream projects to

Comment (grunweg) [reply templates/contribute/style.md:574]: How about splitting that sentence --- something like this?
```
Any publicly exposed theorems and definitions that are being removed should be gracefully transitioned
by keeping the old declaration with a `@[deprecated]` attribute.
This warns downstream projects about the change and gives them the opportunity to adjust before the declarations are deleted.
```

File: templates/contribute/style.md [hunk 4]
```diff
@@ -565,3 +565,29 @@ Documentation strings for declarations are delimited with `/-- -/`.
...
+transition away from deprecated definitions before they are deleted. Renamed theorems
+can use a deprecated `alias` to the new name, while in cases where the theorem is not
```
grunweg, line 576: Renamed definitions can also use this, right? So how about writing something like
```
When a theorem or definition is renamed, please use a deprecated `alias` to the new name, like so:

**Review by grunweg — COMMENTED** [review]
Thanks a lot for creating this PR! I like word-smithing texts and gave this a look: most of these should be quick to address.

File: templates/contribute/style.md [hunk 5]
```diff
@@ -565,3 +565,29 @@ Documentation strings for declarations are delimited with `/-- -/`.
...
+being replaced, the deprecated theorem definition can be marked with an attribute, like so:
+
+```lean4
+theorem new_name : ... := ...
+@[deprecated new_name (since := "YYYY-MM-DD")] alias old_name := new_name
```
grunweg, line 581: Actually, isn't the syntax different? (Here's one random example: https://github.com/leanprover-community/mathlib4/pull/23841/files)
```suggestion
@[deprecated (since := "YYYY-MM-DD")] alias old_name := new_name

File: templates/contribute/style.md [hunk 6]
```diff
@@ -565,3 +565,29 @@ Documentation strings for declarations are delimited with `/-- -/`.
...
+
+### Deprecation
+
+Deleting, renaming, or changing declarations can cause downstreams projects
+that rely on these definitions to fail to compile. Any publicly exposed theorems and
+definitions that are being removed should be gracefully transitioned by keeping the
+old theorem statement with a `@[deprecated]` attribute, to warn downstream projects to
+transition away from deprecated definitions before they are deleted. Renamed theorems
+can use a deprecated `alias` to the new name, while in cases where the theorem is not
+being replaced, the deprecated theorem definition can be marked with an attribute, like so:
```
grunweg, line 577: I wonder what exactly "is not being replaced" means. What do you think about "does not have a direct replacement"? (Perhaps that's just my personal opinion, so I'd like to hear yours.)

grunweg (commented):
... hit submit too soon, my last two comments.

grunweg:
I wonder: would this be a good place to explain mathlib's deprecation policy? (After six months, deprecated declarations may be removed.) Yes, that is scope creep - but that is also not explained yet outside of zulip.

grunweg:
Another idea for scope creep:
- do you think it's useful to say that *named instances* need not be deprecated? (That question also came up occasionally.)

templates/contribute/style.md
```diff
@@ -565,3 +565,29 @@ Documentation strings for declarations are delimited with `/-- -/`.
...
+
+```lean4
+theorem new_name : ... := ...
+@[deprecated new_name (since := "YYYY-MM-DD")] alias old_name := new_name
+
+@[deprecated "This theorem is deprecated in favor of using ... with ..." (since := "YYYY-MM-DD")]
+theorem example_thm ...
+```
+
+The `@[deprecated]` attribute should always include the deprecation date,
```
grunweg, line 587: If I'm reading https://github.com/leanprover/lean4/pull/6112 correctly, the attribute already enforces this. So perhaps
```suggestion
The `@[deprecated]` attribute requires the deprecation date,

**Comment by grunweg** [comment]
There was also https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Deprecation.20policy, trying to gather consensus on a mathlib policy. I never followed through on documenting that... that thread highlighted two more details

- for declarations generated by `to_additive`, please ensure the generated declaration is also tagged with this attribute. For declarations generated by `simps`, this is not strictly required.
- "Cyclic renamings", here's a possible snippet:
`We allow but discourage contributors to rename declaration X to Y and at the same time renaming W to X. In this case, no deprecation attribute is required for X (but it is for W).`

File: templates/contribute/style.md (hunk 5 above)
Comment (vlad902) [reply templates/contribute/style.md:581]: You're right, though according to https://github.com/leanprover/lean4/pull/6112 I thought this should print a warning. It turns out that using an alias (instead of say, a def) seems to suppress the warning, though I'm not sure how.

**Comment by vlad902** [comment]
I hadn't seen that thread, thanks for the feedback! I tried to integrate all of your changes, with some minor modifications of my own to wording where I thought appropriate. Please give it a look and see if you'd like any further changes.

**Comment by grunweg** [comment]
Thanks; I like your new version. There might always be ways to word-smith, but what's here looks good enough to me. I'll cross-post on zulip to get some more eyes on this; let's get this merged!

File: templates/contribute/style.md [hunk 8]
```diff
@@ -565,3 +565,38 @@ Documentation strings for declarations are delimited with `/-- -/`.
...
+The `@[deprecated]` attribute requires the deprecation date, and an alias to the new declaration
+or a string to explain how transition away from the old definition when a new version is no longer
+being provided.
+
+The [`deprecate to`](/mathlib4_docs/Mathlib/Tactic/DeprecateTo.html) command and
+`scripts/add_deprecations.sh` script can help generate alias definitions.
+
+Deprecations for declarations with the `to_additive` attribute should ensure the deprecation is
+also tagged with `to_additive`, while for declarations generated by `simps`, this is not strictly
+required.
```
eric-wieser, line 598: I think the `to_additive` one needs an example, and I'm not sure I believe the claim about `simps`.

grunweg: Agreed, an example for to_additive could help.

jcommelin (approved):
LGTM

templates/contribute/style.md
apnelson1: > The sentence about simps goes back to [this message]([#mathlib4 > Deprecation policy @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Deprecation.20policy/near/436744276)) by @fpvandoorn:
> 
> > I would be fine not mandating deprecations for lemmas generated by simps:
> 
> >  - They are generally not typed manually (though they could appear in simp only calls
> 
> >  - It's very hard to have a lemma generated by simps in your code without the original definition, and the simps lemma should change very predictably. (assuming the only change is renaming the def.)
> 
> 

#23959 deprecated a lot of `to_additive` lemmas, and as discussed in the comments, the process was quite a bit more involved than just adding a tag.

vlad902: @apnelson1 I was not aware of this, thanks for the pointer to that PR! I've removed the bit about `simps` and updated with an appropriate example.

PR #616, opened 2025-04-07
mathlib-overview: support nesting like in undergrad

meithecatte:
Also fixes the spacing of the parentheses for the existing occurrences.


bryangingechen:
Thanks! I checked this locally and it looks good.

PR #619, opened 2025-04-14
feat: add EuroProofNet symposium

grunweg:


grunweg:
While not all of it may be relevant for a typical Lean user, Lean is a diverse group now --- and the announcement mentions the whole event also.

PR #622, opened 2025-04-25
Add Heidelberg Computer-assisted Mathematics Seminar 2025

matematiflo:
Seminar supervised by Judith Ludwig and Florent Schaffhauser.

### DIFF
--- a/data/courses.yaml
+++ b/data/courses.yaml
@@ -50,7 +50,7 @@
   instructor: Judith Ludwig and Florent Schaffhauser
   institution: Heidelberg University, Heidelberg (Germany)
   lean_version: 4
-  tags: ['mathematics', 'formal proof', 'beginner']
+  tags: ['mathematics', 'logic', 'beginner']
   summary: >
     In this seminar, we focus on the formalisation of university-level mathematics in Lean. Our goal is to teach Bachelor students what it means to write a formal proof and have the computer validate it, or even assist us with it.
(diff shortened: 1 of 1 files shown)


data/courses.yaml
```diff
@@ -46,6 +46,15 @@
     The goal of this workshop is to provide an 8h introduction to formal mathematics using Lean. Each 2h session combines a lecture with hands-on computer practice. Practice files for each session are provided and no prior installation of Lean is required.
   website: https://matematiflo.github.io/LeanCompactCourse/
   year: 2025
+- name: Computer-assisted Mathematics
+  instructor: Judith Ludwig and Florent Schaffhauser
+  institution: Heidelberg University, Heidelberg (Germany)
+  lean_version: 4
+  tags: ['mathematics', 'formal proof', 'beginner']
```
robertylewis, line 53:
```suggestion
  tags: ['mathematics', 'logic', 'beginner']
```
or delete this tag if "logic" isn't appropriate? There are no other courses tagged `formal proof`, despite most of them teaching this topic ;)

matematiflo: Change implemented! The 'logic' tag is appropriate. And it would be cool to have a 'formal proof' tag ;)

robertylewis: Thanks! I think the negation of that tag might be more useful -- most of the courses on this list teach formal proof, but I'd be curious which ones don't...

PR #626, opened 2025-05-11
feat: describe naming convention of `≤` and `<` in mathlib

JovanGerb:
This PR describes the new naming convention for `≤` and `<`.

[#mathlib4 > naming convention for ≤ and <](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/naming.20convention.20for.20.E2.89.A4.20and.20.3C/with/514545998)

### DIFF
--- a/templates/contribute/naming.md
+++ b/templates/contribute/naming.md
@@ -131,14 +131,36 @@ When translating the statements of theorems into words, the following dictionary
 
 | symbol | shortcut | name                       | notes                            |
 |--------|----------|----------------------------|----------------------------------|
-| `<`    |          | `lt`                       |                                  |
-| `≤`    | `\le`    | `le`                       |                                  |
+| `<`    |          | `lt` / `gt`                |                                  |
+| `≤`    | `\le`    | `le` / `ge`                |                                  |
 | `⊔`    | `\sup`   | `sup`                      | a binary operator                |
 | `⊓`    | `\inf`   | `inf`                      | a binary operator                |
 | `⨆`    | `\supr`  | `iSup` / `biSup` / `ciSup` | `c` for "conditionally complete" |
 | `⨅`    | `\infi`  | `iInf` / `biInf` / `ciInf` | `c` for "conditionally complete" |
 | `⊥`    | `\bot`   | `bot`                      |                                  |
 | `⊤`    | `\top`   | `top`                      |                                  |
+
+The symbols `≤` and `<` have a special naming convention. In mathlib, we always use `≤` and `<` instead of `≥` and `>`, so we can use both `le`/`lt` and `ge`/`gt` for naming `≤` and `<`. There are a few reasons to use `ge`/`gt`:
+
+1. We use `ge`/`gt` if the arguments to `≤` or `<` appear in different orders. Then `ge`/`gt` indicates that the arguments are swapped.
+2. We use `ge`/`gt` to match the argument order to that of another relation, such as `=` or `≠`.
+3. We use `ge`/`gt` to describe the `≤` or `<` relation with its arguments swapped.
+4. We use `ge`/`gt` if the second argument to `≤` or `<` is 'more variable'.
+```lean
+-- follows rule 1
+#check lt_iff_le_not_ge
+#check LT.lt.not_gt
+
+-- follows rule 2
+#check Eq.ge
+#check ne_of_gt
+
+-- follows rule 3
+#check ge_trans
+
+-- follows rule 4
+#check le_of_forall_gt
+```
 
 ### Dots
 


templates/contribute/naming.md
```diff
@@ -131,15 +131,37 @@ When translating the statements of theorems into words, the following dictionary
...
+| `<`    |          | `lt` / `gt`                |                                  |
+| `≤`    | `\le`    | `le` / `ge`                |                                  |
 | `⊔`    | `\sup`   | `sup`                      | a binary operator                |
 | `⊓`    | `\inf`   | `inf`                      | a binary operator                |
 | `⨆`    | `\supr`  | `iSup` / `biSup` / `ciSup` | `c` for "conditionally complete" |
 | `⨅`    | `\infi`  | `iInf` / `biInf` / `ciInf` | `c` for "conditionally complete" |
 | `⊥`    | `\bot`   | `bot`                      |                                  |
 | `⊤`    | `\top`   | `top`                      |                                  |
 
+The symbols `≤` and `<` have a special naming convention. In mathlib, we always use `≤` and `<` instead of `≥` and `>`, so we can use both `le`/`lt` and `ge`/`gt` for naming `≤` and `<`. There are a few reasons to use `ge`/`gt`:
```
b-mehta, line 143:
```suggestion
The symbols `≤` and `<` have a special naming convention. In mathlib, we almost always use `≤` and `<` instead of `≥` and `>`, so we can use both `le`/`lt` and `ge`/`gt` for naming `≤` and `<`. There are a few reasons to use `ge`/`gt`:
```
to account for things like `\forall x > 0`

templates/contribute/naming.md
```diff
@@ -131,15 +131,37 @@ When translating the statements of theorems into words, the following dictionary
...
+
+-- follows rule 2
+#check Eq.ge
+#check ne_of_gt
+
+-- follows rule 3
+#check ge_trans
+
+-- follows rule 4
+#check le_of_forall_gt
```
b-mehta, line 162: imo this is much more readable if you include the types of these declarations here

JovanGerb: I followed the style of the rest of the document. Would you suggest we change all the examples? By adding a comment with the output of `#check`?

templates/contribute/naming.md
JovanGerb: Indeed. Although I'd also like to suggest changing that situation: `∀ x > 0,` can be notation for `∀ x, 0 < x →`.

templates/contribute/naming.md
b-mehta: The examples after this one don't need changing, because those `#check` aren't explicitly making a point about the type of the lemma. On the other hand, the ones before this (lines 32 to 70) do include the types of the declarations too; my suggestion is to mimic those.

bryangingechen:
This looks reasonable to me. 

Pinging @eric-wieser @YaelDillies in case they have comments.

templates/contribute/naming.md
```diff
@@ -131,15 +131,37 @@ When translating the statements of theorems into words, the following dictionary
...
+| `<`    |          | `lt` / `gt`                |                                  |
+| `≤`    | `\le`    | `le` / `ge`                |                                  |
 | `⊔`    | `\sup`   | `sup`                      | a binary operator                |
 | `⊓`    | `\inf`   | `inf`                      | a binary operator                |
 | `⨆`    | `\supr`  | `iSup` / `biSup` / `ciSup` | `c` for "conditionally complete" |
 | `⨅`    | `\infi`  | `iInf` / `biInf` / `ciInf` | `c` for "conditionally complete" |
 | `⊥`    | `\bot`   | `bot`                      |                                  |
 | `⊤`    | `\top`   | `top`                      |                                  |
 
+The symbols `≤` and `<` have a special naming convention. In mathlib, we almost always use `≤` and `<` instead of `≥` and `>`, so we can use both `le`/`lt` and `ge`/`gt` for naming `≤` and `<`. There are a few reasons to use `ge`/`gt`:
```
YaelDillies, line 143:
```suggestion
The symbols `≤` and `<` have a special naming convention.
In mathlib, we almost always use `≤` and `<` instead of `≥` and `>`, so we can use both `le`/`lt` and `ge`/`gt` for naming `≤` and `<`.
There are a few reasons to use `ge`/`gt`:
```

templates/contribute/naming.md
```diff
@@ -131,15 +131,37 @@ When translating the statements of theorems into words, the following dictionary
...
+
+1. We use `ge`/`gt` if the arguments to `≤` or `<` appear in different orders. Then `ge`/`gt` indicates that the arguments are swapped.
+2. We use `ge`/`gt` to match the argument order to that of another relation, such as `=` or `≠`.
```
YaelDillies, line 146:
```suggestion
2. We use `ge`/`gt` to match the argument order of another relation, such as `=` or `≠`.
```

b-mehta (approved):
Looks good to me

templates/contribute/naming.md
YaelDillies, line 146: Isn't the second rule a generalisation of the first?
```suggestion
1. We use `ge`/`gt` to match the argument order to that of another relation, such as `≤`, `<`, `=` or `≠`.
```

YaelDillies (commented):
The four rules you give are certainly correct, but isn't the following rule encapsulating a few of them?
> We use `ge`/`gt` when the corresponding `le`/`lt` is already taken by an analogous lemma with arguments to some `≤`/`<` swapped.

JovanGerb:
There are generally 2 options when we encounter two similar lemmas that would get the same name if only using `le`/`lt`. Either use `ge`/`gt` to disambiguate, or use ' at the end of the name.

@eric-wieser was confused by some cases where I went with the first approach
```
theorem gt_or_eq_of_ge : a ≤ b → a < b ∨ b = a
```
 [#mathlib4 > naming convention for ≤ and < @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/naming.20convention.20for.20.E2.89.A4.20and.20.3C/near/514550839)

So it's not entirely clear to me what to do in this case, and how to define the rules around this.

YaelDillies:
I personally think it's `lt_or_eq_of_le' : a ≤ b → a < b ∨ b = a`, but `gt_or_eq_of_ge : b ≤ a → b < a ∨ a = b`.

b-mehta:
Those are the same lemma up to alpha-equivalence, right? So to a user who wants to find that lemma, they'd have to guess which variable names the original author used?

YaelDillies:
Yes I agree. My messsage was slightly tongue-in-cheek, but all I wanted to say is that both names are valid interpretations, therefore we should have them both

JovanGerb:
Right, so how far do you want to go with the name duplication? Should we have both forms of
1. `ge_trans`/`le_trans'`
2. `Eq.ge`/`Eq.le'`
3. `ne_of_gt`/`ne_of_lt'`?

YaelDillies:
Yes, I think that doesn't hurt

JovanGerb:
What about
```
le_imp_eq_iff_le_imp_ge' : (a ≤ b → b = a) ↔ (a ≤ b → b ≤ a)
```
Should we duplicate it as
```
ge_imp_eq_iff_ge_imp_le : (a ≤ b → b = a) ↔ (a ≤ b → b ≤ a)
```

Also, does your alpha-non-equivalence mean you don't want to use `alias` for these alpha-equivalent lemmas?

YaelDillies:
No, don't duplicate that one. It is a technical lemma not meant for widespread consumption

bryangingechen:
I'm going to merge this for now as it certainly provides guidance where there was none before, and it looks like we agree that the rules here are a good start. Thanks!

Maybe we can have a follow-up Zulip discussion / PR about when to add aliases with different names.

PR #628, opened 2025-05-17
feat: add recent clarification to community guidelines

grunweg:
Clarify the community guidelines by including the recent annoucnement [on zulip](https://leanprover.zulipchat.com/#narrow/channel/113486-announce/topic/Community.20Policies/near/514574311) in the written text. (This was always implicit policy, but it helps to make it more explicit.)

Specify some further examples of unwelcome behaviour that is disrespectful or disruptive, including using the community to complete your coursework, use of sock puppets and AI abuse.
There may be further clarifications in the future, but the above is already part of the maintainers' consensus.

### DIFF
--- a/data/community.md
+++ b/data/community.md
@@ -83,6 +83,10 @@ These guidelines apply to the
 [Lean Zulip chat](https://leanprover.zulipchat.com/)
 and the [leanprover-community GitHub organization](https://github.com/leanprover-community/).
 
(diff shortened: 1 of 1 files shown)


grunweg:
I have added the words "to clarify the above" (to make some transition from the previous paragraph to the new one). Everything else is copied verbatim from zulip.

grunweg:
CC @fpvandoorn

PR #630, opened 2025-05-22
Add course and update gitignore

mattrobball:
We add a course from Fall 2024 on cryptography to `courses.yaml` and add `.DS_Store` to the `.gitignore` file.

PR #631, opened 2025-05-22
Revert "Add course and update gitignore (#630)"

mattrobball:
This reverts commit 01272cc9ce978d2066645a982512a984bcc8637f. There is a CI error. I will investigate further later.


adomani:
I wonder what went wrong, but let's revert now: thanks!

data/courses.yaml
```diff
@@ -701,23 +701,3 @@
     Monads, CIC, Subtyping, Tactics, Logic, Sets, Computability, Termination,
     Metaprogramming, Macros, Elaboration, Type Unification, and relevant works
     in other proof assistants.
-- name: Introduction to Cryptography
-  instructor: Matthew Robert Ballard
-  institution: University of South Carolina
-  website: https://587.f24.matthewrobertballard.com
-  tages: ['computer science', 'mathematics', 'cryptography']
```
bryangingechen, line 708: The error seems to be here: this `tages` should have been `tags`

adomani: Good catch!

PR #634, opened 2025-05-23
Add course and update gitignore

mattrobball:
This adds a course to the `courses.yaml` and updates the `.gitignore` to include `.DS_Store`. It also strips some trailing whitespace from `courses.yaml`.


bryangingechen:
Thanks! This time it should work!

PR #664, opened 2025-06-28
feat: show how to skip steps 1 to 3 of the git guide with gh

Garmelon:


Garmelon:
I have a nagging feeling that the structure of the one-time setup section can be improved, but I couldn't think of anything substantially better, so I just added a new small section. It's at the top so you notice it immediately instead of after already performing the first few steps.

PR #666, opened 2025-07-01
feat: add warning about default values of structure fields

kbuzzard:
A common gotcha on the Zulip is people using `foo := bar` in structure fields to mean "`foo` is defined to be equal to `bar`" rather than understanding that this is a default value.

Also a couple of unrelated minor changes.

### DIFF
--- a/templates/extras/pitfalls.md
+++ b/templates/extras/pitfalls.md
@@ -34,6 +34,7 @@ If you find a mistake in this document or would like something added, please wri
 - [Non-terminal simp](#non-terminal-simp)
 - [Ignoring warnings](#ignoring-warnings)
 - [Ambiguous unicode characters](#ambiguous-unicode-characters)
+- [Default values in structure fields](#default-values-in-structure-fields)
 
 ## Automatic implicit parameters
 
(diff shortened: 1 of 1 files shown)


templates/extras/pitfalls.md
```diff
@@ -65,7 +66,7 @@ theorem my_theorem : ∃ (a b : ℕ), a ≠ b := sorry
 example : False := Exists.elim (my_theorem (ℕ := False)) (fun f _ ↦ f)
 ```
 
-Automatic implicit parameters are enabled globally by default, but they are disabled if you are working on Mathlib.
+Automatic implicit parameters are enabled globally by default.
```
kbuzzard, line 69: I'm right in claiming that this is no longer the case, right?

templates/extras/pitfalls.md
```diff
@@ -730,7 +731,7 @@ But computationally, `Float` follows the IEEE 754 *binary64* format, which is li
 This means that to prove anything meaningful about `Float`, you have to use `native_decide`, which is a risky tactic (see the section in this document on `native_decide`).
 
 Additionally, because of the behavior of floating point numbers, `0.1 + 0.2 == 0.3` evaluates to `false`.
-(See https://0.30000000000000004.com/ for an explanation.)
+(See [https://0.30000000000000004.com/] for an explanation.)
```
kbuzzard, line 734: Any reason to not make this into a clickable link?

templates/extras/pitfalls.md
```diff
@@ -871,3 +872,35 @@ Character | Unicode Name | How to type | Usage in Lean
...
+If you really want to have something like this, try the following instead:
+
+```
+structure foo : Type where
+
+def foo.n : Nat := 37
+
+def X : foo where
+
+#eval X.n -- 42
```
kbuzzard, line 905:
```suggestion
#eval X.n -- 37
```

kbuzzard: (changing the 42s back to 37s for consistency -- "if you really want to have something like this" and then show them how to have exactly the thing they were trying to do in the failing code.

PR #678, opened 2025-08-30
Update courses.yaml with Lean Tutorial by Wuyang Chen (Simon Fraser University)

chenwydj:


grunweg:
Side note: I'm sure @PatrickMassot is interested in constructive feedback on MIL (such as on the repository https://leanprover-community.github.io/mathematics_in_lean/).
There may be decisions to be made, and sometimes there are different target audiences, but perhaps MIL can take inspiration from what you did.

grunweg:
Sure, thanks for sharing. My impression is that you have edited these chapters to a quite different style than MIL currently has. (I could be wrong, but I don't think such a large rewrite is likely to be adopted in MIL. Just to give an example, I prefer the old way --- thought as not a maintainer, this means essentially nothing.)

If you have specific suggestions that could fit within the current way of writing MIL, filing an issue on that repo could be a good course of action!

adomani:
Thanks!

PR #680, opened 2025-09-09
temporal distribution of commits by type (feat, chore, etc.)

bryangingechen:
We add a new pair of stacked bar charts to the mathlib_stats page which show the number of commits each month / year grouped by commit type (e.g. feat, chore, etc.).

committypes

Follow-up to https://github.com/leanprover-community/mathlib_stats/pull/13. Originally requested by @ocfnash in the maintainers channel.

### DIFF
--- a/templates/mathlib_stats.html
+++ b/templates/mathlib_stats.html
@@ -32,16 +32,11 @@
 
 <canvas id="by_year"></canvas>
 
-<details>
-
-  <summary>Temporal distribution of commits by type (<code>feat</code>, <code>chore</code>, etc.)
-  </summary>
+<h3>Temporal distribution of commits by type (<code>feat</code>, <code>chore</code>, etc.)</h3>
 
   <canvas id="by_labels_by_year_month"></canvas>
 
   <canvas id="by_labels_by_year"></canvas>
-
-</details>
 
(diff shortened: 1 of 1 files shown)


bryangingechen:
There is a slight bug that I haven't bothered tracking down yet. It turns out you can click the labels in the legend (e.g. the colored boxes and their labels near the top of the charts) to toggle displaying their corresponding data. Unfortunately toggling the very first label ("mathlib3: feat") causes the ordering of the mathlib3 and mathlib4 bars to swap positions. I'll leave this as a todo for now since it might require digging into Chart.js to fix.

adomani:
I don't really follow most of the code, but I was wondering if it was possible to also show, maybe with a toggle, the same data, but weighted by number of changed lines, rather than by number of PRs.

Of course, this opens the question of what to do with `+x/-y`: maybe weighting is as `|x-y|` is reasonable?

Anyway, I'm just mentioning it here, since it seems relevant, but, even if it is doable, it can definitely be left for a separate PR!

bryangingechen:
That should be possible if we make [`gitstats.py` in the mathlib_stats](https://github.com/leanprover-community/mathlib_stats/blob/master/gitstats.py) repo generate appropriate data points for us.

PR #686, opened 2025-09-13
Update courses.yaml

sinhp:
adding ""Logic II : Computability, Set Theory, and Model Theory" at Stockholm.


bryangingechen:
Thanks!

PR #829, opened 2026-04-23
feat: a how-to guide for understanding and using `congr()`

Vierkantor:
This PR adds a page explaining the `congr()` term elaborator included with Mathlib. I got a request to explain this a few weeks ago since it is more powerful than appears on first glance, and has some interesting syntactical features that mathematicians (and everyone who isn't a huge Lisp fan) might need some explanations for.

I've also added the new page to the documentation list.

### DIFF
--- a/templates/extras/congr.md
+++ b/templates/extras/congr.md
@@ -13,6 +13,8 @@ the `congr()` elaborator does this for you with less syntactical noise.
 Basic example usage:
 
 ```lean
+import Mathlib.Tactic.TermCongr
+
 example (a b : Nat) (h : a = b) : 4 * (37 + a) = 4 * (37 + b) :=
   congr(4 * (37 + $h))
 ```
--- a/templates/extras/congr.md
+++ b/templates/extras/congr.md
@@ -1,0 +1,196 @@
+# How to use congr()
+
+`congr()` is a term elaborator using *congruence* to build proofs of equality, `HEq` and `Iff`.
+The prototypical congruence lemma is called [`congr`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=congr#doc),
+stating that two expressions are equal (here, two function applications) when the components are equal (here, the functions and their arguments).
+```lean
+theorem congr {f₁ f₂ : α → β} {a₁ a₂ : α} (h₁ : f₁ = f₂) (h₂ : a₁ = a₂) : f₁ a₁ = f₂ a₂
+```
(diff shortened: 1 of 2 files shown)


templates/extras/congr.md
```diff
@@ -0,0 +1,198 @@
...
+  congr(if $h then 1 else 0)
+
+example (a b : Nat) (h : a = b) [NeZero a] [NeZero b] : (37 : Fin a) ≍ (37 : Fin b) :=
+  congr((37 : Fin $h))
+```
+
+## Syntax
+
+Like other elaborators and macros, the parentheses are always attached to `congr()` without a space in between.
+`congr(...)` means something different from `congr (...)`: the first is the term elaborator we are discussing, while the second is the [`congr`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=congr#doc) lemma from the library applied to the arguent `(...)`.
```
grunweg:
```suggestion
`congr(...)` means something different from `congr (...)`: the first is the term elaborator we are discussing, while the second is the [`congr`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=congr#doc) lemma from the library applied to the argument `(...)`.
```

templates/extras/congr.md
```diff
@@ -0,0 +1,198 @@
...
+```lean
+example (a b : Nat) [NeZero a] [NeZero b] (x : Fin a) (y : Fin b) (h : x ≍ y) :
+    37 + x ≍ 37 + y :=
+  congr(37 + $h) -- Error: could not generate congruence between `a` and `b`.
+```
+
+We'd need to construct the proof of `a = b` ourselves and insert it into the expression in the right place to make `congr()` work:
+
+```lean
+example (a b : Nat) [NeZero a] [NeZero b] (x : Fin a) (y : Fin b) (h : x ≍ y) :
```
grunweg, line 189: Perhaps mention the necessary import?
```suggestion
-- import Mathlib.Data.Fin.Embedding to make this work
example (a b : Nat) [NeZero a] [NeZero b] (x : Fin a) (y : Fin b) (h : x ≍ y) :
```

grunweg (commented):
This is great, thanks a lot! Two minor comments.

templates/extras/congr.md
Vierkantor: Nicely spotted!

grunweg:
Thanks again!

PR #835, opened 2026-05-04
chore: update tags and branches page

bryangingechen:
- mention that `stable` follows `v4.X.Y` tags (not just `v4.X.0`)
- update various Zulip links
- add a sentence about `nightly-testing-green`


bryangingechen:
We may want to hold off on merging this until the split mentioned at [#nightly-testing > proposal to split the channel](https://leanprover.zulipchat.com/#narrow/channel/428973-nightly-testing/topic/proposal.20to.20split.20the.20channel/with/592731743) goes through.

PR #839, opened 2026-05-06
Add Values page

ocfnash:
Part of the motivation for these changes is to have a stable link which we can share with AI autoformalisers who erroneously (but understandably) think "Lean says it's true so I'm finished" is the whole story.


chrisflav:
Could you elaborate a bit on how these "missions and values" were drafted? Is this the consensus of the maintainer team or is this your personal suggestion?

grunweg:
This was discussed in the maintainers stream, received broad support (and some feedback) there and no objections. I wouldn't call this a done decision, though: feedback is still welcome.

chrisflav:
The first version contained a "mission statement", which I was (independent of its content) quite surprised by to see without any context. The new version, which only contains the values is much less controversial I think.

chrisflav (commented):
I can't make inline comments at the moment because of the github issue, so let me write some thoughts here:

- In the introductory paragraph for "Implications for PR review" I would link to the review guide. Generally, there is quite a bit of overlap here. Maybe you can comment on what the main purpose of this section is? Is it "give a summary of our review process to outsiders, in particular non-formalists"?
- I would put the "Are the definitions correct" paragraph much higher, maybe after "Is the author a human".
- "instance diamonds": It seems to me that the target audience of this text includes non-formalisers. For them, the term "diamond" will not be known. There are also some other technical terms like `set_option maxHeartbeats` etc.

templates/contribute/values.md
SnirBroshi, line 1: These are great, though I think this is still missing something like "we want good quality proofs", perhaps "0-sorry is a starting point, not an end".
The existing "Maintainability" / "Are the proofs human-readable?" / "Is the code style good?" are technically about this, but I don't think they capture the vibe of "we pride ourselves with good quality code".

ocfnash: > 0-sorry is a starting point, not an end

Thanks I like this phrase and have integrated it.

ocfnash:
> In the introductory paragraph for "Implications for PR review" I would link to the review guide.

Good idea; I did this in https://github.com/leanprover-community/leanprover-community.github.io/pull/839/commits/0aeae75eebd1b36cabb9fb886cb4ffd9339f06ab

> Maybe you can comment on what the main purpose of this section is? Is it "give a summary of our review process to outsiders, in particular non-formalists"?

Yes, the audience here is meant to be non-specialists and thus parts of this content are non-technical versions of remarks in the "Review Guide". Part of the motivation here is to have a link that can we can provide to people who have used AI autoformalisers to help them understand why "Lean says it's true so I've finished" is not the full picture.

> I would put the "Are the definitions correct" paragraph much higher, maybe after "Is the author a human".

Agreed; done.

> "instance diamonds": It seems to me that the target audience of this text includes non-formalisers. For them, the term "diamond" will not be known. There are also some other technical terms like `set_option maxHeartbeats` etc.

Yes, I decided to bend my own rules a bit and include these technical terms. I think it is worth being a bit flexible and allowing this but taking your remarks into account, I have reodered the subsections so that the paragraphs with these more technical terms come last.

templates/contribute/values.md
```diff
@@ -0,0 +1,125 @@
...
+
+We try hard to avoid adding code which is slow to elaborate or to typecheck. If code uses `set_option` to change the default settings of values like `maxHeartbeats` or `synthInstance.maxHeartbeats` then usually something is wrong. Moreover we wish for code to be sufficiently robust that we can expect future code, developed on top of it, will not encounter performance issues.
+
+Lean includes a suite of tools for profiling code. In addition, commenting with `!bench` on a PR provides valuable performance information.
+
+### Data-bearing typeclass instances creating diamonds?
+
+Any new data-bearing typeclass `instance`s are studied during review to increase confidence that diamonds are not being created. These can be difficult to spot and careful thought is needed.
+
+### Are automation annotations in place?
```
faenuccio, line 113: Is it possibile to add links to grind, simp, to_additive and the likes?

templates/contribute/values.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Accessibility for non-formalists
+
+We desire for Mathlib to be accessible to people who do not have experience with computer code beyond LaTeX. Moreover we wish to achieve such accessibility without compromising on the values above. This is a difficult task and much work remains to be done.
+
+Partly for the sake of accessibility, we require human-readable comments (aka "doc strings") on all definitions and we encourage thoughtful code comments in general, especially in longer proofs. Such code comments are often most useful when they use natural language or LaTeX rather than Lean code. We also believe that key to accessibility is the creation of independent artifacts (tutorials, phrasebooks, teaching courses, ...) which build on top of Mathlib and give side-by-side demonstrations of important concepts.
```
faenuccio: Alongside courses,tutorials, etc.. why not mentioning journal papers?

templates/contribute/values.md
ocfnash: Done!

templates/contribute/values.md
ocfnash: Done!

PR #845, opened 2026-05-22
fix(docs): directly link to the tactic list in the menu sidebar

grunweg:
The current link points to the mathlib manual, which points to the tactic list. Skip this level of indirection.


Vierkantor:
Thanks 🎉

PR #851, opened 2026-05-25
feat: document the linebreak policy

YaelDillies:


chrisflav:
I am a bit surprised about the "exceptions" part (and the very quick merge), I think that for example @kim-em has been advocating to put line breaks between all declarations (to make processing files by regex a bit easier).

faenuccio:
Oh, I'm sorry! But given that we've got many instances of this in the library it seems difficult for me to imagine enforcing a policy that we blatantly disregard (so mentioning the exception seems more like a witness of the state of the art). But of course if @kim-em thinks otherwise, I'm happy to revert or to correct the PR myself.

PR #852, opened 2026-05-25
feat: document the `@[to_additive]`/`@[to_dual]` policy

YaelDillies:


templates/contribute/style.md
```diff
@@ -311,6 +311,12 @@ theorem Ordinal.sub_eq_zero_iff_le {a b : Ordinal} : a - b = 0 ↔ a ≤ b :=
    fun h => by rwa [← Ordinal.le_zero, sub_le, add_zero]⟩
 ```
 
+The `@[to_additive]` and `@[to_dual]` attributes are pieces of automation
+that respectively generate additive versions of algebraic multiplicative
+statements and dualised versions of order theoretic statements.
+Where applicable, these attributes should be used
```
faenuccio, line 317: Can you add some link to the relevant descriptions of these tactics, as well as some examples (perhaps also some negative one, where they don't work)?

JovanGerb:
`to_dual` also applies to category theoretic statements, not just order theoretic ones.

PR #853, opened 2026-05-25
feat: document the `@[expose] public section` policy

YaelDillies:


faenuccio:
It seems to me that it is a bit early for enforcing a policy about @expose, and indeed you don't really help new contributors by saying that something is done rarely, and something else in encouraged. I'm not blaming you, nor the wording, I just think that it reflects the current state of the art.

YaelDillies:
I am happy for us to shelve this PR for now if that's preferable

grunweg:
I think this is a better course of action for now. But let's revisit this once we have a clearer idea about `expose` usage!

PR #846, opened 2026-05-22
feat: split the contribute to mathlib page in two

grunweg:
The first page explains what and why to contribute to mathlib (such as, what kind of contributions are welcome, style-only contributions and mathlib's AI policy); the other page covers the how to contribute. Previously, these were concatenated on a single page, which was more confusing.
Update the menu entries on the sidebar accordingly, and rename the entry for the how to page to reflect its broader scope.

PR #843, opened 2026-05-12
`LLM-generated` can be added with comments

SnirBroshi:
leanprover-community/mathlib4#39264 added support for this.

I'm not sure about the phrasing, suggestions welcome

PR #855, opened 2026-05-28
fix: `LLM-generated` can be added with comments

SnirBroshi:
This was added in #843 but lost during #846


grunweg:
Sorry for missing that, thanks for the fix! (Did you, by chance, find other things to fix?)

PR #857, opened 2026-05-29
chore: disambiguate maintainers as mathlib maintainers

grunweg:
This is already what the short description uses: make it very clear.

### DIFF
--- a/data/teams.yaml
+++ b/data/teams.yaml
@@ -18,7 +18,7 @@
     - Sebastian Ullrich
     - Eric Wieser
   use_biography: True
-- name: Mathlib maintainer team
+- name: Mathlib maintainers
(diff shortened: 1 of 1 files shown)


data/teams.yaml
```diff
@@ -18,7 +18,7 @@
     - Sebastian Ullrich
     - Eric Wieser
   use_biography: True
-- name: Maintainer team
+- name: Mathlib maintainer team
```
faenuccio: Why isn't it "Mathlib maintainers team" ?

grunweg: Good question! Shall we go with "Mathlib maintainers" as the synthesis of our ideas?

faenuccio: Yes, I think it is the best option. Clear, unambiguous.

grunweg:
```suggestion
- name: Mathlib maintainers
```

PR #876, opened 2026-06-19
style: fix typo in PR convention doc

luigi-massacci:
The last example did not adhere to the conventions.


grunweg (commented):
I had also noticed that - thanks for PRing this before I got to it!

PR #885, opened 2026-06-25
feat: document the conjunction/disjunction convention

YaelDillies:


grunweg:
Thanks. This one seems uncontroversial and definitely existing practice (so I suggest merging this in 24 hours).
Can you add a sentence of motivation to the document?

grunweg:
Related, but a separate topic: we typically don't want conclusions involving a \forall, \exists or \and either. Would you like to document this also? (I don't mind if in this or a separate PR.)

templates/contribute/style.md
```diff
@@ -401,6 +401,33 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+
+example {i} (hV : V i) : W := ...
+```
+
+Similarly, the return type of a lemma should not be a conjunction.
+One should instead prove two lemmas.
+It is acceptable to prove the two lemmas from a conjunction lemma if that reduces code duplication.
+
+Finally, an existential result can be turned using choice into a definition along with a lemma
+about that definition. Whether this is a sensible change to make highly depends on the situation.
```
joneugster:
```suggestion
Finally, an existential result can be turned, using choice, into a definition along with a lemma about that definition. Whether this is a sensible change to make depends highly on the situation.
```

some missing commas and wrong word order.

joneugster (approved):

templates/contribute/style.md
YaelDillies: This might be a German-English difference?

b-mehta: I agree that the commas should be added. To me, "depends highly" and "highly depends" are both clear though, but "depends highly" feels more natural.

b-mehta:
I posted on zulip, but `Exists` and `Or` don't follow existing practice uniformly. (`And` does though, that one I'm fully in favour of!). 

> we typically don't want conclusions involving a \forall, \exists or \and either.

However I strongly disagree that we don't want conclusions containing `Exists`!

templates/contribute/style.md
YaelDillies:
```suggestion
Finally, an existential result can be turned, using choice, into a definition along with a lemma
about that definition. Whether this is a sensible change to make depends highly on the situation.
```

YaelDillies:
Do you agree with my phrasing about `Exists` in conclusions?

There is indeed a valid use case for `Or` or `Exists` in hypotheses but it is an overarching minority of all places where it could be used and I couldn't find an easy way to convey the rule. Do you have a suggestion here?

b-mehta:
> Do you agree with my phrasing about `Exists` in conclusions?

I agree that your phrasing about `Exists` in conclusions is correct, although I'm not sure if it's helpful? It's essentially saying "we don't have a rule", which isn't so different from not mentioning it at all. 

> There is indeed a valid use case for Or or Exists in hypotheses but it is an overarching minority of all places where it could be used and I couldn't find an easy way to convey the rule. Do you have a suggestion here?

Yeah I agree completely, it's the minority and it's not clear to me how to convey what the rule actually is. Let me try to propose some phrasing as a suggestion, and see what you think.

templates/contribute/style.md
```diff
@@ -401,6 +401,33 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+
+example (hS : S) : U := ...
+example (hT : T) : U := ...
+
+example {i} (hV : V i) : W := ...
+```
+
+Similarly, the return type of a lemma should not be a conjunction.
+One should instead prove two lemmas.
+It is acceptable to prove the two lemmas from a conjunction lemma if that reduces code duplication.
```
b-mehta:
```suggestion
Similarly, the return type of a lemma should not be a conjunction, and one should instead prove two lemmas.
It is acceptable to prove the two lemmas from a private conjunction lemma if that reduces code duplication.
```

templates/contribute/style.md
```diff
@@ -401,6 +401,33 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+
+Finally, an existential result can be turned, using choice, into a definition along with a lemma
+about that definition. Whether this is a sensible change to make depends highly on the situation.
```
b-mehta:
```suggestion
Finally, an existential result can be turned, using choice, into a definition along with a lemma about that definition. 
Whether this is a sensible change to make depends highly on the situation.
```
optional, but this follows the line-breaking in the rest of your PR!

templates/contribute/style.md
```diff
@@ -401,6 +401,33 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+example (hV : ∃ i, V i) : W := ...
+
+-- ...do
+example (hP : P) (hQ : Q) : R := ...
+
+example (hS : S) : U := ...
+example (hT : T) : U := ...
+
+example {i} (hV : V i) : W := ...
+```
```
b-mehta:
````suggestion
Hypotheses should not be conjunctions, as this typically makes the lemma harder to use:
```lean
-- Instead of
example (hPQ : P ∧ Q) : R := ...
-- do
example (hP : P) (hQ : Q) : R := ...
```

In the majority of situations, hypotheses should not be disjunctions or existentials, for the same reason:
```lean
-- Instead of
example (hST : S ∨ T) : U := ...

example (hV : ∃ i, V i) : W := ...
-- do
example (hS : S) : U := ...
example (hT : T) : U := ...

example {i} (hV : V i) : W := ...
```
````

Comment (b-mehta) [reply templates/contribute/style.md]: My intention is to make it clearer that the conjunction rule is a rule, and the disjunction/existential is a guideline.

File: templates/contribute/style.md [hunk 5]
```diff
@@ -401,6 +401,56 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+Hypotheses should not be conjunctions, as this typically makes the lemma harder to use:
+```lean
+-- Instead of
+example (hPQ : P ∧ Q) : R := ...
+-- do
+example (hP : P) (hQ : Q) : R := ...
+```
+
+Similarly, the return type of a lemma should not be a conjunction, and one should instead prove two lemmas.
+It is acceptable to prove the two lemmas from a private conjunction lemma if that reduces code duplication.
```
b-mehta:
```suggestion
It is acceptable, however, to prove the two lemmas from a private conjunction lemma if that reduces code duplication.
```
Minor but this makes it clear that this is an allowed exception, rather than an explanation or clarification of the previous line.

templates/contribute/style.md
```diff
@@ -401,6 +401,56 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+In the majority of situations, hypotheses should not be disjunctions, for the same reason:
+```lean
+-- Instead of
+example (hST : S ∨ T) : U := ...
+-- do
+example (hS : S) : U := ...
+example (hT : T) : U := ...
+```
+
+Exceptions can be made when abiding to this rule would result in many lemmas with similar function:
```
b-mehta:
```suggestion
Exceptions can be made when abiding to this rule would result in many very similar lemmas:
```
optional, but I think I like the phrasing in your example more

templates/contribute/style.md
```diff
@@ -401,6 +401,56 @@ is also valid. In mathlib the pretty printer displays `↦`, and we slightly pre
...
+```
+
+A similar transformation could be made to existential hypotheses:
+```lean
+-- Instead of
+example (hV : ∃ i, V i) : W := ...
+-- one could do
+example {i} (hV : V i) : W := ...
+```
+This is mostly discouraged as providing `i` explicitly can be harder than proving its existence.
```
b-mehta: I think "mostly discouraged" might be a bit strong, eg there's a bunch of lemmas in mathlib which take `{i} (hV : V i)` and I don't want someone modifying them all to take existentials. Can we use phrasing similar to the conclusion one below - where we mention the option but don't enforce one way or the other?

b-mehta: (In particular, this is now the negation of what it was earlier, which is what Michael called uncontroversial!)

b-mehta (approved):
I'm happy with this PR now, perhaps @faenuccio or @grunweg would like to take a look?

faenuccio:
I'm also happy with this, thanks.

bors merge

YaelDillies:
No bors here 😁 You have to merge the good old way

faenuccio:
Oh sure, we' re not on `mathlib`, I'm sorry :rofl: I opened the doc to check one thing about this PR and got trapped thinking I was dealing with a purely mathlib PR... thanks for catching this!

PR #886, opened 2026-06-25
feat: document the function application naming convention

YaelDillies:


templates/contribute/naming.md
```diff
@@ -554,3 +554,33 @@ theorem Continuous.mul (hf : Continuous f) (hg : Continuous g) : Continuous (f *
 Both theorems deserve tagging with the `fun_prop` attribute.
 
 The same goes for addition, subtraction, negation, powers and compositions of functions.
+
+### Function application
+
+Function application `f x` is denoted in three ways in lemma names:
+1. It is omitted when `f` is a concrete function/morphism and `x` is a concrete argument,
```
b-mehta: "It" here is ambiguous, my first reading was that mentioning `f` is omitted! Perhaps "The application is omitted"?
Also, "concrete" isn't explained anywhere in this document, right? Don't forget that readers of this document might not know what you mean by that.

b-mehta (commented):
I agree with the spirit of this convention, but I think the current phrasing is not very clear.

templates/contribute/naming.md
YaelDillies: How about now?

templates/contribute/naming.md
```diff
@@ -554,3 +554,34 @@ theorem Continuous.mul (hf : Continuous f) (hg : Continuous g) : Continuous (f *
...
+  or more generally if `f` is a concrete function in several arguments
+  and at least one of them is concrete.
+  This case makes up the overwhelming majority of all occurrences of function application:
+  Remember that almost everything in Lean is a function applied to some argument!
+2. The application is denoted by the `_apply` suffix when `f` is a concrete function/morphism
+  and `x` is a free variable. Many such lemmas are generated by `simps` on a concrete morphism.
+3. The application is denoted by the `map_` prefix when `f` is a free variable
+  or the coercion to function thereof, and `x` is a concrete argument.
+  Most lemmas of this form are about properties of a generic morphism.
+
```
b-mehta, line 572: Your change is an improvement, but I think it's still a little ambiguous whether or not `a + b` is concrete or not... I wonder if it's easier to explain in terms of what's *named*. In all your examples, concrete and "is part of the name" coincide exactly

YaelDillies: Hmm, I am not sure. Beginners sometimes name their lemmas after free variables, no?

b-mehta: Fair point, but I still think your phrasing doesn't sufficiently explain what happens to `a + b`

YaelDillies: I've attempted something. I personally find it more confusing now, but you might disagree

templates/contribute/naming.md
```diff
@@ -554,3 +554,34 @@ theorem Continuous.mul (hf : Continuous f) (hg : Continuous g) : Continuous (f *
...
+```lean
+-- A concrete function with a concrete argument
+lemma Rat.inv_zero : (0 : Rat)⁻¹ = 0 := ...
+
+-- A concrete binary function with one concrete argument
+lemma Int.mul_zero (a : Int) : a * 0 = 0 := ...
```
eric-wieser, line 578: Here I think we should show zero_mul too

templates/contribute/naming.md
YaelDillies: Also note that I show below `*` as an example of a concrete function, so maybe there was no need to clarify in the first place?

b-mehta: Actually that was the reason it confused me, from your description I didn't think you intended it to be a concrete function, but your example included it.

YaelDillies: Aha ok 😆 I hope now the phrasing is clear enough?

templates/contribute/naming.md
```diff
@@ -554,3 +554,35 @@ theorem Continuous.mul (hf : Continuous f) (hg : Continuous g) : Continuous (f *
 Both theorems deserve tagging with the `fun_prop` attribute.
 
 The same goes for addition, subtraction, negation, powers and compositions of functions.
+
+### Function application
+
+Function application `f x` is denoted in three ways in lemma names:
+1. The application is omitted when `f` is a concrete function/morphism
```
b-mehta, line 561:
```suggestion
1. Mention of the application is omitted when `f` is a concrete function/morphism
```
I read this again and it took me a few passes to figure out what the wording meant, even though I understand the convention here!

b-mehta:
Thanks!

PR #891, opened 2026-07-24
Make the lean3 warning even more obvious, and strikethrough write-access docs

SnirBroshi:
Followup to #377

This uses the same red color from the [mathlib3 docs](https://leanprover-community.github.io/mathlib_docs/) (just `alert-danger` isn't as prominent), and makes the banner stick to the top to prevent people with section links from missing it.

Also strikethrough the write-access note in the [mathlib3 contribution guide](https://leanprover-community.github.io/lean3/contribute/index.html), and add a note to clarify it is outdated.

Before:
image
After:
image

[#mathlib4 > github write access](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/github.20write.20access/with/612455049)

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -47,15 +47,15 @@
 <div id="mainbox">
 
   <main>
-    <div class="alert alert-warning">
+    <div class="alert alert-danger" style="background: #b00020; color: white; position: sticky; top: 0;">
       <p>
-      This webpage is about Lean 3, which is <strong>effectively obsolete</strong>;
+      This webpage is about Lean 3, which is <u>effectively obsolete</u>;
       the community has migrated to Lean 4.
       </p>
       {% if name is defined %}
       <p>
       Jump to the
-      <a href="https://leanprover-community.github.io/{% if name.endswith('.md') %}{{ name[:-3] }}.html{% else %}{{ name }}{% endif %}">corresponding page</a>
+      <a href="https://leanprover-community.github.io/{% if name.endswith('.md') %}{{ name[:-3] }}.html{% else %}{{ name }}{% endif %}" style="color: #88ccff;">corresponding page</a>
       on the main Lean 4 website.
       </p>
       {% endif %}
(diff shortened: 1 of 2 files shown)


bryangingechen (approved):

grunweg:
I like making the warning more obvious, so clear support for the first change.

I wonder if the second one is really necessary: if people were actually reading the page, with the large red banner on top, perhaps they would realize this is outdated? (After all, the "outdated" is technically wrong for the mathlib3 policy, and only true because the page is outdated. If anything, we should make the *entire* page strikethrough.)
Overall, the second change makes me somewhat sad: I'd rather drop it, or strike through the entire page.

SnirBroshi:
I agree it's odd to both leave the old site up for historical reasons and scrub the write-access part, but since some people read everything through an LLM I'm not sure the glaring red banner would stop them from asking for write-access.

The strikethrough and "please ignore" is a compromise in order to preserve the old page as much as possible.
Removing that part would also work, but strikethrough on everything without the "please ignore" wouldn't work against LLM readers, as I'm not sure they can see the strikethrough (or the banner).

thorimur:
Perhaps the note should make clear it's from the future and was not in the original. Something like *(Note as of 2026: this information is outdated and should be ignored; mathlib4 now uses a typical fork model.)*

I agree it's rather odd, but it may help the google search overview situation. :/

grunweg:
That's a nice solution! I'm sad this is necessary these days, but I can live with the phrasing. I'm happy with this PR being merged now.

grunweg:
No further comments in a day plus epsilon - let's merge this!

SnirBroshi:
Thanks! Could you deploy it?

Looks like the [CI failed](https://github.com/leanprover-community/leanprover-community.github.io/actions/runs/30620345780/job/91123169220) because it's pretty old (on the `lean3` branch).

bryangingechen:
This is live now: https://leanprover-community.github.io/lean3/contribute/index.html

PR #896, opened 2026-08-13
feat: mention #min_imports in imports style documentation

JadAbouHawili:


templates/contribute/style.md
```diff
@@ -50,7 +50,10 @@ a description of the contents.
 Put the `module` keyword on its own line right after the header, skip a line,
 then group all `public import`s together, skip another line,
 then group all `import`s together,
-Try to keep the imports alphabetical within each block of imports.
+Try to keep the imports alphabetical within each block of imports
+and avoid specifying imports which are already transitively imported by another import.
```
grunweg, line 54: This would be a new policy, so deserves a careful look. In this case, there is no general agreement about this being a good change - so please revert this sentence. (What is true is that one shouldn't randomly add transitively implied imports - but that might be understood in general, hence shouldn't be said here.)

templates/contribute/style.md
```diff
@@ -50,7 +50,10 @@ a description of the contents.
...
+We recommend using `#min_imports` at the end of the file which would output
```
grunweg: I wouldn't say "recommend" (that's too strong) --- more that it's a possibility.

That said, I'm not sure if this document is the best place for such general advice. This style guide is describing rules to follow, not giving advice to users. (I don't see an obvious other place to put it either, unfortunately.)

grunweg (commented):
Thanks for trying to keep the docs updated. I applaud the idea, but I'm not sure this particular change is useful. I have explained this in more detail below.

templates/contribute/style.md
SnirBroshi: Note that currently `#min_imports` is not mentioned anywhere on the site, which seems wrong.

JadAbouHawili: changed the wording to not say "recommend" , just to say that using `#min_imports` would output something that satisfies these conditions.

I think that mentioning `#min_imports` here (or some other docs) is an improvement

templates/contribute/style.md
JadAbouHawili: > (What is true is that one shouldn't randomly add transitively implied imports )

I agree , which is why you should "try" "to avoid specifying imports which are already transitively imported by another import" , which is the current wording.

A policy would be something like:
"Put the module keyword on its own line right after the header, skip a line, ..."
"The file header should contain copyright information, ..."

JadAbouHawili: I can instead say "and try to avoid specifying imports ... " to make the point clear

JadAbouHawili:
@grunweg what do you think? Anything else that needs to be changed?

PR #904, opened 2026-08-24
Add Julian to the moderator team.

Julian:
(Following https://github.com/leanprover-community/moderation_docs/blob/main/Welcome_message_template.md )


data/teams.yaml
```diff
@@ -55,7 +55,6 @@
     - Adam Topaz
     - Eric Wieser
   use_biography: True
-  url: maintainers
```
Julian, line 58: (This seems like an unrelated bug, the key is defined twice here. Found via a YAML linter which runs by default when I edit the file, might be worth running one in CI?)

Julian:
Whoops, thanks (for the title fix, I was multitasking).

grunweg:
Thanks, and welcome onboard!

PR #905, opened 2026-08-26
feat(contribute/commit.md): allow commit subject to start with uppercase letter

joneugster:
suggestion from a reviewer discussion: [#mathlib reviewers > Check PR title](https://leanprover.zulipchat.com/#narrow/channel/345428-mathlib-reviewers/topic/Check.20PR.20title/with/604858902)

### DIFF
--- a/templates/contribute/commit.md
+++ b/templates/contribute/commit.md
@@ -42,7 +42,7 @@ For instance, it could be
 `<subject>` has the following constraints:
 
 - use imperative, present tense: "change" not "changed" nor "changes"
-- do not capitalize the first letter
+- do not capitalize the first letter, unless the subject starts with a proper noun (e.g. "Kan extensions")
 - no dot(.) at the end
 
 `<body>` has the following constraints:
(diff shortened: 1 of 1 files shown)


SnirBroshi:
Counterpoint: prefixing "Kan extensions" with a verb would probably make it better, e.g. "define Kan extensions" or "expand API for Kan extensions", etc.
Forcing a verb is a win for the linter IMO, not a false-positive.

Perhaps "Euler's theorem" is a better example for why uppercase is fine, since "prove Euler's theorem" / "add Euler's theorem" sound less cool, but IMO the lint is worth this small trade-off.

Do you have other examples which aren't improved by a lowercase prefix?

joneugster:
For context, this came from somebody in the reviewer channel above complaining about the linter being inconvenient in several instances. I made this PR after a 7:0 vote (I myself was neither initiated nor participated in that vote) in favour of disabling the linter. My point was that the linter should not be disabled unless the underlying convention is adjusted first. 

So here is the PR which would do that, and I'd leave it up to the people being troubled by the linter to move this forward :)

One example raised was https://github.com/leanprover-community/mathlib4/pull/43121

grunweg:
Let me explicitly highlight that this PR should be merged to mathlib-landing instead! (There'll be a proper announcement etc. soon, just to prevent human error until then.)

PR #907, opened 2026-08-26
feat: remove suggestion to always self-introduce

grunweg:
I'm not sure we always want to encourage this (and it's mostly not followed in practice anyway). Accompanies https://github.com/leanprover-community/mathlib4/pull/42901.


joneugster (approved):
thanks, I support this change :)

Vierkantor:
Since mathlib4#42901 is merged, something of this shape seems like obviously worth merging too. I am happy with the phrasing here. Shall I merge this unless there is an objection in 24 hours?

Vierkantor:
No objection in 24 hours, so let's get this merged!

PR #909, opened 2026-09-03
feat: add documentation about cross-reference tags

luigi-massacci:
To the best of my knowledge, these weren't really documented anywhere, despite the facts that we now have a half dozen databases and hundreds of tags (mostly stacks but nevermind...). This seemed like the most sensible location.


bryangingechen:
Thanks!

PR #913, opened 2026-09-10
feat(style): clarify the scope of the deprecation policy

grunweg:
Structure fields, and simps-generated lemmas are publicly reachable: explain that, since I was wondering about it.
In contrast, auto-generated declarations are not reachable.

Mention the Lean declarations diff which helps with this.


b-mehta (approved):
I think this is a positive change.

Vierkantor:
I like this change and I see two positive responses: merging 🎉

PR #914, opened 2026-09-11
feat: naming conventions for meta interfaces

thorimur:
This PR adds basic, uncontroversial naming conventions for meta interfaces such as tactics, attributes, and commands.

### DIFF
--- a/templates/contribute/naming.md
+++ b/templates/contribute/naming.md
@@ -529,11 +529,13 @@ The same goes for addition, subtraction, negation, powers and compositions of fu
 
 ## Tactics, elaborators, and other meta interfaces
 
-The following list refers to the naming of surface syntax the user will write (e.g. the tactic `norm_num`), not declarations per se.
+The following list refers to the naming of surface syntax the user will write (e.g. the tactic `norm_num`), not declarations per se. In general, `lower_snake_case` is the correct casing for new syntax.
 
 - Tactics should use `lower_snake_case`, e.g. `norm_num`.
 - Attributes should use `lower_snake_case`, e.g. `@[to_additive]`.
-- Commands should use `lower_snake_case`, e.g. `deprecated_module`, and prefixed by `#` if and only if they are temporary or test commands (i.e. not to be committed to mathlib) such as `#check`. (`#adaptation_note` is an exception to this, but is in some sense a "longer-term temporary command".)
-- Non-notation term elaborators should use `lower_snake_case`, and may often be followed by `%`, e.g. `fast_instance%`. Note that this convention does not necessarily apply to elaborators which attempt to "mimic" other declarations or functions while providing additional functionality, such as `MDiff[_]`.
+- Commands should use `lower_snake_case`, e.g. `deprecated_module`, and should be prefixed by `#` if and only if they are temporary or test commands (i.e. not to be committed to mathlib) such as `#check`.
+  - `#adaptation_note` is an exception to this, but is in some sense a "long-term temporary command".
+- Non-notation term elaborators should use `lower_snake_case`, and may often be followed by `%`, e.g. `fast_instance%`.
+  - This convention does not necessarily apply to elaborators which attempt to "mimic" other declarations or functions while providing additional elaboration-time functionality, such as `MDiffAt[_]`.
 
-Note that declaration names in meta code should follow the typical naming conventions.
+Declaration names in meta code should follow the typical naming conventions.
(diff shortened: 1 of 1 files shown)


JovanGerb:
Thanks, the explanation looks good to me.

joneugster (approved):

bryangingechen:
I think we can merge this if you fix the conflict!

templates/contribute/naming.md
```diff
@@ -526,3 +526,16 @@ theorem Continuous.mul (hf : Continuous f) (hg : Continuous g) : Continuous (f *
...
 
 The same goes for addition, subtraction, negation, powers and compositions of functions.
+
+## Tactics, elaborators, and other meta interfaces
+
+The following list refers to the naming of surface syntax the user will write (e.g. the tactic `norm_num`), not declarations per se. In general, `lower_snake_case` is the correct casing for new syntax.
+
+- Tactics should use `lower_snake_case`, e.g. `norm_num`.
+- Attributes should use `lower_snake_case`, e.g. `@[to_additive]`.
+- Commands should use `lower_snake_case`, e.g. `deprecated_module`, and should be prefixed by `#` if and only if they are temporary or test commands (i.e. not to be committed to mathlib) such as `#check`.
```
SnirBroshi: This sounds good to me, but does this mean we should change `#whats_new in` to `#whats_new_in` (and `#defeq_abuse`, `#count_heartbeats`, etc, perhaps even `@[pibase topology]` to `@[pibase_topology]`, but not `@[to_dual existing]`), or should `in` be mentioned as an exception?

Also some syntaxes have multiple atomic parts, such as `"apply" foo "at" bar`, and although currently all of them are not commands and the second atom is a single word, what should be the convention for such cases? e.g. `#diff def foo := 3 compare_with example := trivial` or could the underscore be a space? or perhaps it should also start with `#` even though it's part of the same command, since it'll always appear on a separate line (it follows a `command`)

thorimur: Oh, good point, I guess I should clarify that this convention only applies to *tokens* in the syntax! (And the prefixing only to the first token.) Then all of these are fine, and separating tokens by a space or using underscores is not determined by these rules :)

JovanGerb: I think it might even be worth saying explicitly that commands that take in another command should end with the `in` keyword.

thorimur: Hmm, I'm not sure. I can imagine that we don't want to think of `#diff` as being "in" the following command? Also `#find_home for`, as another example 😁

In any case I wanted this PR to just be the uncontroversial conventions instead of all of them! :)

templates/contribute/naming.md
```diff
@@ -608,3 +608,18 @@ To disambiguate, we suffix the series 3 name with `₀`.
...
 lemma inv_eq_self₀ {K : Type*} [DivisionRing K] {a : K} : a⁻¹ = a ↔ a = -1 ∨ a = 0 ∨ a = 1
 ```
+
+## Tactics, elaborators, and other meta interfaces
+
+The following list refers to the naming of surface syntax the user will write (e.g. the tactic `norm_num`), not declarations per se. In general, `lower_snake_case` is the correct casing for tokens in new syntax.
+
+- Tactics should use `lower_snake_case`, e.g. `norm_num`.
+- Attributes should use `lower_snake_case`, e.g. `@[to_additive]`.
+- Commands should use `lower_snake_case`, e.g. `deprecated_module`, and the first token should be prefixed by `#` if and only if it is a temporary or test commands (i.e. not to be committed to mathlib) such as `#check`.
```
grunweg:
```suggestion
- Commands should use `lower_snake_case`, e.g. `deprecated_module`, and the first token should be prefixed by `#` if and only if it is a temporary or test command (i.e. not to be committed to mathlib) such as `#check`.
```

grunweg:
Thanks, this looks good to me! In this case, the earth moved under you (so to say): this webpages has moved to mathlib.org. Can you file a PR with the same change to mathlib.org (at https://github.com/leanprover-community/mathlib-landing)? Otherwise, I can also do so.

thorimur:
Sure! Off topic, but should I also migrate #919?

thorimur:
Also, do I just target `main`? I notice there's no `lean4` branch.

grunweg:
Yes, just targetting main is good. And migrating #919 also would be great, thank you!

thorimur:
Migrated to leanprover-community/mathlib-landing#5.

PR #916, opened 2026-09-14
feat: replace old lean-pr-testing-* docs with downstream-lean4 docs

Garmelon:
The old `lean-pr-testing-*` infrastructure has been turned off and replaced by the `leanprover/downstream-lean4` repository.


Vierkantor (approved):
Thanks for the update!

PR #919, opened 2026-09-21
feat: module system quickstart guide

thorimur:
A short guide describing the basics of how the module system works (but not why).

---

Note: a longer explainer is in the works, for those interested. But I feel a shorter version is probably useful for most.


thorimur:
~~Migrated to leanprover-community/mathlib-landing#6~~ Migrated back here :)

thorimur:
By the way @grunweg, is this website still going to be used for resources relevant to the broader community? If so, this PR actually might be relevant to anyone using the module system 👀

grunweg:
Actually, you're right: this resource is better placed here. Sorry for the false alarm.

Vierkantor (commented):
From a first (surface-level) readthrough my impression is good, nice work!

Before I forget to ask: could you please add this to the `documentation.yml` overview? (Not sure if there's any other place you can think of to link to this page.)

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+`import A` puts the public scope of `A` into the _private_ scope of the current module.
+
+`public import A` puts the public scope of `A` into the _public_ scope of the current module.
+
+This also means that downstream imports of the current module, in receiving the current module's public scope, will also necessarily receive the public scope of `A` along with it.
+
+<img src="img/module-system/vis-import-of-public-import.svg" alt="A diagram depicting an import following a public import by showing three module pill diagrams in a row, with an arrow from A's public dot to B's public dot, and an arrow from B's public dot to C's private dot." class="diagram"/>
+
+(The arrows in these diagrams may be composed.)
```
Vierkantor: Not immediately obvious to me what this meant on first reading. However, it is correct after thinking about it and I can't think of a good alternative phrasing.

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+Typical `import`s, `public` or not, only request the public scope from the upstream module. The upstream private scope stays inaccessible unless you write `import all`, which should generally be avoided. (Even then, the upstream private scope can only be imported into the private scope; private data can never be made public.)
+
+As an escape hatch, `import all A` will request the private scope of `A` as well as the public one, and put it in the target module's private scope. However, note that `public import all` is not allowed: the private scope of `A` cannot ever be lifted into the public scope. `public import A` and `import all A` (on separate lines) will import the public scope into the public scope and the private scope into the private scope.
+
+<img src="img/module-system/vis-table-with-composites.svg" alt="A table of pairs of module pill diagrams, showing an import (A's public dot connected to B's private dot); a public import (A's public dot connected to B's public dot); an import all (A's private dot connected to B's private dot); and a public import with an import all (A's public dot connected to B's public dot, and likewise for private dots)." class="diagram"/>
```
Vierkantor: I suppose the grey arrows here are the composition mentioned above? Maybe a label for the meaning of the grey arrow could be helpful. Or they could be made black without really affecting the meaning, no?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+An ordinary declaration has several pieces of data associated with it:
+
+- `name : Name`
+- `type : Expr`
+- (optional) `value : Expr`
+- (optional) `code : IR`
+
+In some sense "the constant" is simply its name. Its type, value, and so on are merely data _associated_ with the constant.
+
+Each constant has a _visibility_, i.e. which scope it is in (public or private).
```
Vierkantor: This appears to draw a distinction between "declaration" and "constant", which I'm not sure I can follow: is the difference that constants have a visibility additionally? [Lean's documentation style guide](https://github.com/leanprover/lean4/blob/master/doc/style.md#constants) suggests we should call these things "(defined) constants", and `decl` as a name for constants only really appears in the compiler as far as I can tell.

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+While a constant's current visibility depends on how it is imported (publicly or privately), it also has an _intrinsic_ visibility, namely which scope of its originating module it lives in.
+
+Note: this intrinsic visibility is apparent in the `name`: private constants are subject to mangling, and their name will have an autogenerated prefix like `_private.Foo.Bar.0.myFoo`, where `Foo.Bar` is the original module of `private def myFoo`. Though the actual visibility may change depending on how its imported, the name of a constant is immutable, and never changes no matter where or how it is imported.
+
+The `type` of a constant always has the same visibility as the constant itself.
+
+The `value` of a constant cannot have a greater visibility than the constant itself (it makes no sense to have a public value for a private constant, for example), but it may have a _narrower_ visibility. Namely, the `type` of a constant may be public while the value is private.
+
+Having a private value is in fact the default situation for a `public def`, and the _only_ situation for a `public theorem`. (The `value` is the theorem's proof, and proofs are always private.)
```
Vierkantor: What exactly counts as a proof in "proofs are always private": I know that the value of a theorem or the contents of a `by` block count, but some term-level values that live in `Prop` seem to be exceptions?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+When importing a public constant with a private `value` in an ordinary fashion (i.e. without `import all`), the `value` data will simply not be loaded downstream, and thus will be completely inaccessible.
+
+#### `@[expose]`
+
+However, `public` definitions may, if we so choose, have public `value`s as well. We may raise the intrinsic visibility of a `public` definition's `value` to `public` by attaching `@[expose]` to the definition.
+
+Note that when the value of a constant is available at a given visibility, we also have a _rewrite rule_ that lets us _unfold_ the constant to its value during definitional equality checks. When exactly we are allowed to unfold an `@[expose]`d constant is controlled by the definition's transparency settings.
```
Vierkantor: Does this hold for definitional equality checks in the kernel too? (I assume it does, but the kernel does not respect transparency settings.)

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+As such, `@[expose]` is required for unfolding. Conversely, without `@[expose]`, unfolding is not possible downstream even in principle under ordinary imports (since the necessary private `value` data is not available). Transparency settings (`@[reducible]`, `@[implicit_reducible` , and so on) are thus _only_ meaningful downstream on `@[expose]`d definitions. Otherwise, they only affect unfolding in private locations in the same file.
+
+Note: even without `@[expose]`, the private `value` data of a definition is still available in the module in which it was created, but only in the private scope. This accordingly means that we may be able to unfold even a non-exposed definition in private locations in the module in which it was defined (such as proof bodies) without being able to unfold it at _any_ point downstream (unless we were to `import all`, which would give us exactly that private data; note, even then, we would still only be able to unfold it in private locations).
```
Vierkantor: For clarity, I'd switch the order of these explanations. Something like:

A constant can be only be unfolded if its value is in scope, namely:

* in the private scope of the module where it is defined
* in the private scope of a module that `import`s `all`
* if `@[expose]`d, wherever the constant is in scope, whether public or private.

Only in these conditions the transparency settings are meaningful, otherwise unfolding is blocked in all cases.

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+To recap:
+
+- `def`:
+    - `type`: same visibility as overall constant
+    - `value`: private by default, can be made public (when applicable) via `@[expose]`
+- `theorem`:
+    - `type`: same visibility as overall constant
+    - `value` (i.e. proof): always private
+
```
Vierkantor: Do we want to mention (`@[expose]` and/or `public`) `section` here? Or is that more for the reference manual?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+The `meta` keyword controls what happens to executable code. Specifically, executable code may be either available for execution during the elaboration of the current module, in which case we call it _meta_ code, or not, in which case we call it _runtime_ code. This state ("meta" or "runtime") is called the _phase_ of the code.
+
+Meta code cannot be used in runtime code, and vice versa.
+
+`meta import A` _lifts_ all of the runtime code reachable from `A` to the meta phase. That is, we are now able to use runtime code from `A` in meta code we write in our current module. (Meta code from `A` stays meta.)
+
+`public meta import A` does the same, but also provides the same meta-lifted runtime code to downstream modules of the current module.
+
+Note that any code from `A` which is _already_ intrinsically marked as meta will still be made available and remain meta under an ordinary import. For example, macros are intrinsically meta, and may be obtained by downstream modules via an ordinary non-meta `import`. I.e., intrinsically meta code is unaffected by `meta import`; the `meta` keyword on imports only lifts the phase of runtime code, and does nothing else.
```
Vierkantor: "intrinsically marked as meta": sounds a bit complicated, when it basically comes down to "has the `meta` keyword in its declaration" right? (If not, I'd at least give it as a concrete example.)

templates/extras/module-system-quickstart.md
Vierkantor: Additionally, does the same hold for types, since they don't have a runtime representation either?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+The `meta` keyword _only_ affects the phase of executable code, and otherwise is ignored. So, a `public meta import` functions as a `public import` for the purpose of type-theoretic terms. As such, a `theorem` (which has no executable code) is totally insensitive to whether its dependencies are meta or runtime (both are accepted).
+
+Because compilation erases proofs and types, this means that the phases of constants used by proofs and types even in definitions which _do_ produce compiled code are ignored when it comes to enforcing the phase distinction.
```
Vierkantor: This seems exactly one paragraph too early to mention the compiler. "Because computation ignores proofs and types" maybe?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Code
+
+Declarations in Lean, separately from their role as terms in a type theory, are also _compiled_ into code (Lean IR), which may then either be compiled further to machine code, or may be executed directly by the Lean interpreter. As such, each constant is also associated to _code_.
+
+A constant's code has a phase, i.e. may be either meta or runtime.
```
Vierkantor: Maybe:
```suggestion
A constant's code has a phase, i.e. may be either meta or runtime, as explained above.
```

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+However, note that code has a complicated relationship to visibility. The meta code associated with a constant does _not_ behave similarly to the value of the constant, even though it comes from compiling the value. Instead, the `code` of a `public meta def` should be thought of as somehow "public" itself. (Recall that, in contrast, whether the actual `value` is public is controlled by `@[expose]`.) The code of a `public meta def` requires all of its transitive dependencies to be both meta and have publicly-imported modules.
```
Vierkantor:
```suggestion
However, note that code has a complicated relationship to visibility. The meta code associated with a constant does _not_ behave similarly to the value of the constant, even though it comes from compiling the value. Instead, the _code_ of a `public meta def` should be thought of as somehow "public" itself. (Recall that, in contrast, whether the actual _value_ is public is controlled by `@[expose]`.) The code of a `public meta def` requires all of its transitive dependencies to be both meta and have publicly-imported modules.
```
Since `code` and `value` are not used as keywords / identifiers before, make them italic instead of monospace?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+Though, due to a module being publicly available to itself, a `public meta def` is allowed to use private `meta def`s from the same module—as long as _those_ `def`s, in turn, only have dependencies which are meta and have publicly-available originating modules. This leads to situations such as
```
Vierkantor:
```suggestion
Though, due to a module being in scope for itself, a `public meta def` is allowed to use private `meta def`s from the same module—as long as _those_ `def`s, in turn, only have dependencies which are meta and have publicly-available originating modules. This leads to situations such as
```
might be clearer?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+Presentability behaves analogously to visibility with respect to imports: `public (meta) import A` imports all of the presentable code from `A` as presentable code in the target module; a `(meta) import` imports all of the presentable code from A as unpresentable.
+
+Note: if the same code is provided transitively in different ways from different imports, it has the maximum presentability among all different ways it is provided. For example, if `foo` is imported as unpresentable code due to `import A'` and as presentable code due to `public import A''`, it is presentable.
+
+`(public) meta import A` lifts the runtime code from `A` (and its dependencies) to the meta phase, making it presentable or unpresentable depending on whether `public` is or isn't present, respectively.
+
+Code that is already intrinsically `meta` (i.e. marked meta in the source file) is still brought in as meta code under ordinary imports, and is unaffected by the presence of `meta` on the `import`.
```
Vierkantor:
```suggestion
Code that is already intrinsically meta (i.e. marked `meta` in the source file) is still brought in as meta code under ordinary imports, and is unaffected by the presence of `meta` on the `import`.
```
I would flip the order of the backticks here: the first is an English adjective, the second a Lean keyword.

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+In the following diagram, the arrows mean "(transitively) referenceable in", and coloration only shows (redundantly) where the arrows end, to emphasize that runtime IR becomes available in the meta phase under `meta import`s. (These arrows may be composed.)
```
Vierkantor: Might also be worth mentioning that there is intentionally no arrow going into runtime code, since runtime code cannot refer to meta code?

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+-- Succeeds
+meta def foo := bar -- note: `bar` is privately `meta import`ed
+
+-- Fails, due to `bar` in `foo` not being publicly imported
+public meta def fooPublic := foo
+```
+
+The above example succeeds if we instead publicly import the module of `bar`.
+
+We can conceptualize it this way: instead of saying that code is "public" or "private", we use the similar terms "presentable", meaning _valid to use downstream_, or "unpresentable".
```
Vierkantor: Since runtime code does not participate with presentability, I'd make it explicit here as well:
```suggestion
We can conceptualize it this way: instead of saying that meta code is "public" or "private", we use the similar terms "presentable", meaning _valid to use downstream_, or "unpresentable".
```

(And perhaps an extra sentence to already mention here that code in the runtime phase does not participate in the presentability system.)

Vierkantor (approved):
Some more in depth editorial reading. Nicely written and very educational! 🎉 I have only one piece of structural feedback and otherwise relatively minor style suggestions.

templates/extras/module-system-quickstart.md
```diff
@@ -0,0 +1,180 @@
...
+
+<img src="img/module-system/vis-single-module-with-arrow.svg" alt="Two dots labeled 'public' and 'private' enclosed by a pill-shape labeled 'A', with an arrow reaching down from the 'public' dot to the 'private' dot." class="diagram"/>
+
+`import A` puts the public scope of `A` into the _private_ scope of the current module.
+
+`public import A` puts the public scope of `A` into the _public_ scope of the current module.
+
+This also means that downstream imports of the current module, in receiving the current module's public scope, will also necessarily receive the public scope of `A` along with it.
+
+<img src="img/module-system/vis-import-of-public-import.svg" alt="A diagram depicting an import following a public import by showing three module pill diagrams in a row, with an arrow from A's public dot to B's public dot, and an arrow from B's public dot to C's private dot." class="diagram"/>
```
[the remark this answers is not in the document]
thorimur: I know what you mean, and actually I started with arrows in that direction! 😅 The labels were what pushed me this way: `A` is on the "right" of e.g. the `public import A` statement itself. So if we flip them, we wind up with something like
```
A                B
    <diagram>

  public import A
```
and it makes it a bit awkward to say *what* is publicly importing A, as now `B: public import A` further brings us away from the diagram. I guess there's something like `public import A from B`, but space is at a premium...

Maybe a box around it and a corner label? Effectively

```
A                B
    <diagram>

|-----------------B
| public import A |
|-----------------|
```
but that still looks a little weird/bulky, I think. I could prototype! Would you like me to experiment with that, or do you think this makes sense now given this reasoning? 😅

templates/extras/module-system-quickstart.md
thorimur: Hmm, what about: "(Information flows through the arrows in these diagrams transitively, meaning we can chain arrows together to infer new, composite arrows.)" Or a variation on that?

templates/extras/module-system-quickstart.md
thorimur: Right, exactly.

Here's the black version, but I greyed these arrows to show they weren't what was "driving" the flow of info. Hypothetically you could just think about the black ones then infer the gray ones.
image

Do you think we should have either
1. a line above that says something to the effect of "Gray arrows are inferred from chaining together the black arrows where possible."
2. an in-diagram label like

image

3. black arrows?

(I'd lean towards 1 to keep the diagram cleaner, I think, with (2) as my second preference.)

templates/extras/module-system-quickstart.md
thorimur: I use them more or less interchangeably, starting with "declaration" and transitioning to "constant" when I get more precise, since I think "declaration" is the more common name the user might be familiar with (we use "declaration" to refer to, well, the constant in quite a few places in natural language in mathlib/on zulip and in meta code (`decls`)) and connotes the fact that this data is specified/controlled by the user. Afaict, "declaration" is more precisely used to refer to the command written by the user and any data derived from it which is used to specify what constant to add to the environment...though mentally we often think of those as the same thing, or the declaration as including the constant, since they usually occur simultaneously and one subsumes the other. (At least, I *think* people do...)

But anyway, in this specific block I should use "constant" I think, since it gives the appearance of a distinction where there really doesn't need to be any.

templates/extras/module-system-quickstart.md
thorimur: It's a bit weird...

So, here's the reality, and then we can discuss what should be presented 😁

When you're in a `theorem` body or in a `by` block for a `Prop` (which, recall, will be delayed until it has a known type), *elaboration* changes, and the syntax is elaborated in a private context. Meaning that identifiers for private constants resolve successfully, `isDefEq` can unfold private definition bodies, etc.

When you're in e.g. a random proof argument to a type former, let's say in constant's type, elaboration does not know about this or take it into account even if it *does* know about it. So you aren't allowed to resolve identifiers for private constants etc. (Hence the `private` elaborator to allow elaboration to happen at a private visibility :) leanprover-community/mathlib4#42563).

But then lean *also* does its own nested proof abstraction in between elaboration and adding the actual declaration, in public (exposed) definition values:
```
@[expose] public def half (n : Nat) (h : 0 < n) : { m : Nat // m < n } :=
  ⟨n / 2, Nat.div_lt_self h (Nat.lt_succ_self 1)⟩

set_option pp.proofs true
/--
info: @[expose] def half : (n : Nat) → 0 < n → { m // m < n } :=
fun n h => ⟨n / 2, half._proof_1 n h⟩
-/
#guard_msgs in
#print half
```
(It doesn't do this in the type afaik!)

We were still forced to use public identifiers at elaboration time though, even though this eventually becomes private.

And more broadly, proofs always *may as well* be made private even if they aren't always made private, due to proof irrelevance, but that point kind of muddies things.

But more importantly (to the second point) this has nothing to do with the runtime representability. The compiler is totally irrelevant to discussions about the private and public nature of types + proofs :) This may-as-well-be-private fact and the usually-private nature of proofs relies only on *type-theoretic* proof-irrelevance (and how that affects `isDefEq`), not the compiler's proof erasure.

I should probably revise this bit to clarify things. 🤔 (Let me know if you have any sense of what beats are important to hit here! I'll also give it a second attempt on my own.)

templates/extras/module-system-quickstart.md
thorimur: Right, the kernel cannot "see through" a non-exposed def from upstream (assuming we haven't written `import all`, since the data of what the body is is literally just not there. We haven't looked at the olean to which that data is written.

But, does the kernel respect our "working" visibility? 🤔 So, would it unfold it if we were in the non-exposed def's original file (in which its body is privately available), or we had written `import all`, would it have unfolded? I suspect it would unfold. But I'm not sure. I wonder also if it depends on what calls the kernel (e.g. if it's the kernel being called during elaboration, or at the end of a file), and whether this can be controlled when calling it.

Let me get Claude to do an audit, which will at least give us an approximation or hopefully point me to the relevant code...there's probably a way to test as well.

templates/extras/module-system-quickstart.md
thorimur: Hmm, I agree about restructuring here (it's a bit awkward...), but I do want to somehow emphasize that it all follows from what the visibility of the value is, and what visibility you're working at. So the private scope of a module isn't special; the fact that you can unfold an `@[expose]`'d  constant in the private scope of the module where it is defined is simply because the public scope of that module is available!

I'm definitely going to reattempt this bit either way, but could you share more about your reasoning here (what makes things clearer, what the underlying issue is)? (Of course some of these things are ineffable, so "well, I think it's better this way" is totally valid... 😆)

templates/extras/module-system-quickstart.md
thorimur: Oh, good point. I should have a small section on `section` :)

templates/extras/module-system-quickstart.md
thorimur: This is a bit tricky I think, as "in scope" could also mean "in the private scope of"...in general I think "in scope" is a tiny bit dangerous as a concept here, because we all instinctively know what it should mean, and *feel confident* about knowing what it means it, but (if we're learning) we may not necessarily know what really *possesses* the scope, and therefore create a "confidence blind spot" where we don't realize we're relying on something we don't fully understand.

Because really, whether something is "in scope" is *relative* to the scope we're working at, and this scope is not attached to a module per se, as we may change what scope we're working at freely even within the same module. But at the same time, the manner in which we import things for a module determines whether the imported module is "in scope" for *both* of the scopes we might be working at (at once), so there is a sense in which we might think "in scope" is a module-level statement rather than one relative to our present circumstances/elaborator state!

Sorry, that's a tangent on "in scope" as verbiage in general. Do you think this information I've just mentioned should be emphasized somewhere, though?

On the actual comment, I'm not sure what a better wording is, but I'll think about it! It should communicate that the public scope of this module *is present whenever we import it downstream*. (I.e. it's the transitive nature of "publicly available" through imports that's relevant here.)

templates/extras/module-system-quickstart.md
thorimur: They are, way up just under `## Declarations`...but maybe that's too far away?

